PR #43, opened 2016-07-02
Implement optional `options` argument for changeset.save

dustinfarris:
Enable ember-data [record.save(options)](https://github.com/emberjs/data/blob/v2.6.1/addon/-private/system/model/model.js#L704) in order to leverage things like [adapterOptions](https://github.com/emberjs/data/blob/v2.6.1/addon/-private/system/snapshot.js#L34-L39) when saving a changeset that has an ember-data model as its content.


poteto:
Thanks, LGTM!

PR #47, opened 2016-07-05
Fix bug where virtual properties were not being notified when changed

poteto:
This was causing an issue where observing a value on the changeset (e.g.
in the template) would not work.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -341,12 +341,14 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
           this.notifyPropertyChange(ERRORS);
         }
 
+        set(changes, key, value);
         this.notifyPropertyChange(CHANGES);
+        this.notifyPropertyChange(key);
 
-        return set(changes, key, value);
-      } else {
-        return this.addError(key, { value, validation });
+        return value;
       }
+
+      return this.addError(key, { value, validation });
     },
 
     /**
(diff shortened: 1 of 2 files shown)


dustinfarris:
Thanks for this. If you want I can verify before you merge.

dustinfarris:
This does not fix for me.

dustinfarris:
This does not fix for me, I'll try to put together a failing test, but difference in my use-case is use of component wrapper around input and an action to mutate changeset property.  (instead of 2WB in your test)

poteto:
@dustinfarris I updated the integration test, it still passes with a `mut` action

dustinfarris:
still not doing it for me, but maybe user error on my part

my usage is

page template:

```hbs
{{registration-form changeset=(changeset model UserValidations)}}
``

form component:

``hbs
{{paper-input value=changeset.email onChange=(action (mut changeset.email))}}
```

paper-input implementation: https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-input.js#L210
paper-input template: https://github.com/miguelcobain/ember-paper/blob/master/app/templates/components/paper-input.hbs

poteto:
Odd, at this point I'm uncertain if this is still a bug on ember-changeset. Can you use a regular / native input in place of `paper-input` and see if you still get the same issue?

dustinfarris:
@poteto it is working now.  I think ember-changeset-validations was pulling in the old version of ember-changeset.  I upgraded to latest ember-changeset-validations which references latest release of ember-changeset and everything clicked.

Thanks for the quick turnaround on this.

dustinfarris:
Actually, I am noticing the same bug when I pass a validations map to the changeset helper.  Will open an issue there.

PR #50, opened 2016-07-06
Update test to verify property update

### LINKED ISSUE #49: property not notified when using a validator
I am getting the bug fixed in #47 when i pass a validator (either as an action or an ember-changeset-validators validation map).

```hbs
{{#with (changeset model (action "validate") as |changeset|}}
  <h2>{{changeset.email}}</h2>
  <input value={{changeset.email}} oninput={{action (mut changeset.email) value="target.value"}}>
{{/with}}
```

poteto:
Closes #49

### DIFF
--- a/tests/integration/components/changeset-test.js
+++ b/tests/integration/components/changeset-test.js
@@ -89,10 +89,31 @@ test('it can be used with 1 argument', function(assert) {
   assert.notOk(this.$('p:contains("There were one or more errors in your form.")').length, 'does not turn invalid');
 });
 
-test('it updates when set', function(assert) {
+test('it updates when set without a validator', function(assert) {
   this.set('dummyModel', { firstName: 'Jim', lastName: 'Bob' });
   this.render(hbs`
     {{#with (changeset dummyModel) as |changeset|}}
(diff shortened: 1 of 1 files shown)


poteto:
@dustinfarris I have added tests for both with and without a validator function, and it seems to still be updating. I think your issue might be something else.

dustinfarris:
Update the validator in your test to return `false` and see if it still passes.

dustinfarris:
@poteto aha, that would be it then.  so i still have an issue in ember-paper where the paper-input matches the `value` property exactly.

poteto:
Sounds like the issue should be opened on ember-paper then! We have fixed the bug on our end, so I will close #49 once I merge this in.

dustinfarris:
Yes indeed.  Thanks for your support.

miguelcobain:
Isn't it strange to have:
```
> changeset.get('email')
undefined
> changeset.set('email', 'dustin')
"dustin"
> changeset.get('email')
undefined
```
?

The set itself returned the value, but then a get doesn't.
From the perspective of a component, it should always represent `email`. If it is `undefined`, I just show `undefined`?
What am I missing here?

Issue #24, opened 2016-06-11
Think about handling constraints


poteto:
Needs some kind of server interaction and possible adapter implementation

dustinfarris:
I'm thinking something like this (works sort of locally):

```js
// in changeset#save
return content
  .save(options)
  .then(() => this.rollback())
  .catch(response => this.evaluateConstraints(response));


evaluateConstraints(response) {
  let content = get(this, CONTENT);
  let constraintErrors = get(this, CONSTRAINT)(content, response);
  mergeErrorsSomehow(constraintErrors);


// Some user-provided or included function
constraintFn(content, response) {
  // Could potentially inspect response for a server error or something
  // Otherwise if we're using ember-data the errors are stored on the model
  return content.get('errors.messages');  // or maybe iterate through each specific attr
}
```

dustinfarris:
@poteto what do you think of this direction?  Should I start working on a PR or did you have something else in mind?

poteto:
I need to think about it a little more, ideally I would like a constraint to be caught prior to saving. But that would mean your API would need to have some endpoint for eg to check uniqueness for a given email address. And technically you could do that right now with an asynchronous validator

poteto:
The proposed solution is more of a feature to merge server side errors with the changeset, which could be a useful feature. It's not related to constraints in Ecto parlance though

dustinfarris:
Oh? My understanding is that this is precisely what Ecto constraints do.  From their docs:

> Ecto changesets provide both validations and constraints which are ultimately turned into errors in case something goes wrong... constraints can only be checked in a safe way when performing the operation in the database. As a consequence, validations are always checked before constraints. Constraints won’t even be checked in case validations failed.

Basically swap out "database" with "backend API" and that's what I'm trying to accomplish here.

poteto:
Ah yeah I got mixed up, my bad! My concern is that a server side error isn't necessarily always a constraint error, so it would not entirely be correct to call it a "constraint" on the addon's end. For example, having `uniqueConstraint` or `foreignKeyConstraint` methods on the changeset wouldn't be possible as we won't be able to tell from simply receiving the error response from the API. It would be more correct to simply have some method that merges server side errors (whatever they may be) onto the changeset. It might just be naming, but I think it's an important distinction to make. Not sure if I explained it well, let me know if it doesn't!

That API for adding errors could just be:

```js
changeset
  .save()
  .then(/* ... */)
  .catch((errors) => {
[...]

Issue #51, opened 2016-07-06
addError overwrites existing error for key


dustinfarris:
Is this intended behavior?  If so we should mention in the docs.

Also, as an aside, addError is given an array for `validation` message from ember-changeset-validations, but is given a string in the docs and test.  Is this something we care about and/or will care about in future development?

poteto:
Yes it is intentional. Would you like to open a PR for it?

It doesn't really matter what the value for `validation` is, it is dependent on the validation library. In the case of `ember-changeset-validations`, it's always an array. But in ember-changeset it doesn't have to.

PR #52, opened 2016-07-06
Example of handling ember data errors

dustinfarris:
I wanted to automate this originally, but after tinkering with it I think the best short term solution is to document how to do this in your own app—it isn't that hard after all.

The PR looks bigger than it is—my editor stripped whitespace.  Separated into 2 commits.

Ref #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,6 +465,25 @@ export default Controller.extend({
 
 Your action will receive a single POJO containing the `key`, `newValue`, `oldValue` and a one way reference to `changes`.
 
+## Handling Server Errors
+
+When you run `changeset.save()`, under the hood this executes the changeset, and then runs the save method on your original content object, passing its return value back to you.  You are then free to use this result to add additional errors to the changeset if applicable.
+
+For example, if you are using an Ember Data model in your route, saving the changeset will effectively save the model.  If the save rejects, Ember Data will add errors to the model for you.  To copy the model errors over to your changeset, add a handler like this:
+
+```js
+changeset.save()
+  .then(() => { console.log('Success!'); })
+  .catch(() => {
+    let errors = get(this, 'model.errors');
+    errors.forEach(error => {
+      let key = error.attribute;
+      let validation = error.message;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -457,13 +457,34 @@ export default Controller.extend({
...
+
+When you run `changeset.save()`, under the hood this executes the changeset, and then runs the save method on your original content object, passing its return value back to you.  You are then free to use this result to add additional errors to the changeset if applicable.
+
+For example, if you are using Ember Data and would like to add any model attribute errors that may result from the save back to the changeset:
+
+```js
+changeset.save()
+  .then(() => { console.log('Success!'); })
+  .catch(() => {
+    const errors = this.get('model.errors');
```
poteto: `let` please, and use `get(this, 'model.errors');`

README.md
```diff
@@ -457,13 +457,34 @@ export default Controller.extend({
...
+    errors.forEach(error => {
+      let key = error.attribute;
+      let value = changeset.get(key);
```
poteto: I wouldn't pass `value` in here, it wouldn't make any sense since `get` would only get a valid value. Just the `validation` itself is fine

README.md
```diff
@@ -457,13 +457,34 @@ export default Controller.extend({
...
+      let validation = [error.message];
```
poteto: It doesn't have to be an array, see #51 (`ember-changeset` is unopinionated about the validation message and leaves  it up to the validation library to return whatever). I would not wrap this in an array since it might potentially be confusing.

README.md
```diff
@@ -457,13 +457,34 @@ export default Controller.extend({
...
+      changeset.addError(key, { value, validation });
```
poteto: `changeset.addError(key, { validation });` will suffice

dustinfarris:
@poteto this is ready

poteto:
@dustinfarris Needs a rebase

dustinfarris:
Done

PR #55, opened 2016-07-07
Fix `Changeset#get` for properties with blank values

jmbejar:
This commit fixes a problem where changes in the changeset object could
not be retrieved if the value was something like `undefined`, `null`,
`''`, etc.

This issue was not a problem at the moment of executing the changes in
the original object, but it was a problem when you rely on the temporary
changes for other things (e.g external computed properties, template).


poteto:
LGTM, thanks!

PR #61, opened 2016-07-11
Fix issue with `unknownProperty`

### LINKED ISSUE #60: validating single property causes value to rollback to last valid value
To reproduce:

1. Enter invalid value in demo
2. Tab out
3. Input's value rollsback to last valid value

poteto:
Prior to this commit, `changeset.get` would only retrieve the value from
the changes object or the underlying model itself. The problem with this
is that when an input contains an invalid value (let's say this is "foo"
in an email-only input), the `changes` object will only contain the last
valid value, and thus the input resets when attempting to call
`validate`. This commit makes it so that `unknownProperty` will also
fetch the value from the error object, which ensures that
`changeset.get` is in sync with the input.

Closes #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ get(changeset, 'isDirty'); // true
 
 #### `get`
 
-Exactly the same semantics as `Ember.get`. This proxies to the underlying Object, or the changed value if present.
+Exactly the same semantics as `Ember.get`. This proxies first to the error value, the changed value, and finally to the underlying Object.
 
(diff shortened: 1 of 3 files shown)


dustinfarris:
@poteto does this mean that `changeset.addError` should always be sure to include `value`?  Re: your comment here on adding ember-data errors: https://github.com/DockYard/ember-changeset/pull/52#discussion_r69779142

Issue #66, opened 2016-07-14
Conditional validations and fields


ivanvanderbyl:
I have a use case where some form fields are only required if a certain value is selected somewhere else in the form (in this case a checkbox). Upon investigating how validations and changesets work, I can see it's not so trivial to add support for this because not only do you need to not disable the validations, but you also need to ensure that field doesn't get propagated to the model on execute.

Any thoughts on this? Am I missing some not so obvious way to achieve this? Otherwise, I'll have a crack at implementing it soon.

poteto:
Hmm unless I'm misreading it, I think you could do this using ember-changeset-validations with a custom validator. Are you using that add-on for validations? This add-on is only the logic for the changeset itself.

ivanvanderbyl:
Hi @poteto ! I am indeed using `changeset-validations`, but as far as I can tell the individual validations don't receive a reference to the changeset, so there's no way to query the value of the checkbox during validation.

Upon thinking over this further, I don't think it's just disabling the validation that is needed, because in that case you might just return true and assume the attr in valid. What is really needed is excluding that value from the `changeset` and not validating it.

poteto:
@ivanvanderbyl The validator function receives the `changes` object, which is a map of all the changes that have been made. Would that work? Also, this issue is probably more relevant on the `ember-changeset-validations` addon!

https://github.com/DockYard/ember-changeset-validations#writing-your-own-validators

ivanvanderbyl:
Yes, I noticed that, however `changes` alone aren't enough to solve this problem.

Imagine the scenario: You have a form with a conditional section, in this case it collects a bunch of EC2 configuration details like VPC ID, Security Group, AZ, and Subnets. These are all required fields only if the "Use EC2" checkbox is ticked, in which case they must send the default value of the model, which is `null`. 

Now I can easily discard the validations of these properties if `useEc2` is false. However, if these fields change I can't set default value back to `null` from within the `validate` function as it's not very "pure", and I'm trying to only use the `changeset` helper to construct the Changeset.

A solution I'm attempting is adding an additional form helper to `flood-simple-form` which exposes a conditional section, ala `fieldset`, with its own changeset and validations. If the condition is true and it's valid it gets merged into the main changeset, otherwise it's removed. At least, that's the thinking, we'll see how well it works in practice.

poteto:
[...]

PR #74, opened 2016-07-19
Added option to allow merging invalid changesets

ivanvanderbyl:
From discussion in #66

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,7 +396,7 @@ user.get('firstName'); // "Jimmy"
 user.get('lastName'); // "Fallon"
 ```
 
-Changes are merged using similar heuristics to [`Object.assign`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign), such that `B` will replace `A` in `changesetA.merge(changesetB)`. Neither changeset is actually mutated, instead a clean changeset is created from `A` and `B` which contains `errors` from `A` and `B` where `errors` on `A` do not include valid keys on `B`, and valid keys on `A` do not include errors on `B`.
+When merging, the changeset being merged in takes precedence. For example: if `B` contains errors, the errors from `B` will be merged, and the invalid attributes on `B` will not be merged from `A` even if they're valid on `A`. Likewise, if `A` contains errors on attributes which are valid on `B`, the value from `B` will be used and the errors for those keys will not be merged from `A`. This is to prevent the new changeset from containing `changes` and `errors` on the same keys, which is a guarantee of a normal changeset.
 
 Note that both changesets `A` and `B` are not destroyed by the merge, so you might want to call `destroy()` on them to avoid memory leaks.
 
--- a/README.md
+++ b/README.md
@@ -386,9 +386,7 @@ Merges 2 changesets and returns a new changeset with the same underlying content
 let changesetA = new Changeset(user, validatorFn);
 let changesetB = new Changeset(user, validatorFn);
 changesetA.set('firstName', 'Jim');
-changesetA.get('errors') // [{key: 'firstName', value: 'Jim', validations: ['Firstname cannot be Jim, sorry Jim']}]
 changesetB.set('firstName', 'Jimmy');
-changesetB.get('errors') // []
 changesetB.set('lastName', 'Fallon');
 let changesetC = changesetA.merge(changesetB);
 changesetC.execute();
--- a/README.md
+++ b/README.md
@@ -396,7 +396,7 @@ user.get('firstName'); // "Jimmy"
 user.get('lastName'); // "Fallon"
 ```
 
-Changes are merged using the same heuristics as Object.assign, such that `changesetB` will replace `changesetA` in `changesetA.merge(changesetB)`. Neither changeset is actually mutated, instead a clean changeset is created from `changesetA` and `changesetB` which contains `errors` from `changesetA` and `changesetB` where `errors` on `changesetA` do not include valid keys on `changesetB`, and valid keys on `changesetA` do not include errors on `changesetB`.
+Changes are merged using similar heuristics to [`Object.assign`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign), such that `changesetB` will replace `changesetA` in `changesetA.merge(changesetB)`. Neither changeset is actually mutated, instead a clean changeset is created from `changesetA` and `changesetB` which contains `errors` from `changesetA` and `changesetB` where `errors` on `changesetA` do not include valid keys on `changesetB`, and valid keys on `changesetA` do not include errors on `changesetB`.
 
 Note that both changesets `A` and `B` are not destroyed by the merge, so you might want to call `destroy()` on them to avoid memory leaks.
 
--- a/README.md
+++ b/README.md
@@ -396,8 +396,6 @@ user.get('firstName'); // "Jimmy"
 user.get('lastName'); // "Fallon"
 ```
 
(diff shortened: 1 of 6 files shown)


poteto:
I was actually thinking merge would merge changesets regardless of validity by default. I think it was too heavy handed in the beginning - the user should be able to decide since the changeset can only be executed if it is valid. What do you think?

ivanvanderbyl:
Yeah, totally, I'm happy with that. PR updating...

ivanvanderbyl:
Hmm so some thoughts on merging logic:

Let's assume we have the following (taken from readme):

```js
let changesetA = new Changeset(user, validatorFn);
let changesetB = new Changeset(user, validatorFn);
changesetA.set('firstName', 'Jim');
changesetA.get('errors') // [{key: "firstName", value: "Jim", validations: ["Firstname cannot be Jim"]}]
changesetB.set('firstName', 'Jimmy');
changesetB.set('lastName', 'Fallon');
let changesetC = changesetA.merge(changesetB);
```

Now if we merge `changesetA` which has an invalid `firstName`, and `changesetB`, which does not have an invalid `firstName`, what should happen?

1. We pick the one with the valid attr and keep that,
2. We don't merge the attr because one of them has errors,
3. We assume merging order indicates intention and `B.firstName` replaces `A.firstName`.

Thoughts?

ivanvanderbyl:
Essentially what I'm getting at is, after merging you shouldn't end up with changes on `firstName` and errors on `firstName`

poteto:
Yup agreed, I think it should work the same way `Object.assign` does -- the last object should override any previous value from previous objects

addon/index.js
```diff
@@ -224,6 +227,14 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
       newChangeset[CHANGES] = mergedChanges;
       newChangeset.notifyPropertyChange(CHANGES);
 
+      if (allowInvalid) {
+        let errorsA = get(this, ERRORS);
+        let errorsB = get(changeset, ERRORS);
+        let mergedErrors = pureAssign(errorsA, errorsB);
+        newChangeset[ERRORS] = mergedErrors;
+        newChangeset.notifyPropertyChange(ERRORS);
```
poteto: Sidenote: use `this._notifyVirtualProperties();`

ivanvanderbyl:
I think that makes sense, I'll reject keys from errors on A which are present changes of B. That actually would make logical sense in my form because each section is merged from top to bottom.

ivanvanderbyl:
@poteto I think this solves it ^^

README.md
```diff
@@ -380,20 +380,24 @@ changeset.rollback(); // returns changeset
...
 #### `merge`
 
-Merges 2 valid changesets and returns a new changeset with the same underlying content and validator as the origin. Both changesets must point to the same underlying object. For example:
+Merges 2 changesets and returns a new changeset with the same underlying content and validator as the origin. Both changesets must point to the same underlying object. For example:
 
 ```js
 let changesetA = new Changeset(user, validatorFn);
 let changesetB = new Changeset(user, validatorFn);
 changesetA.set('firstName', 'Jim');
+changesetA.get('errors') // [{key: 'firstName', value: 'Jim', validations: ['Firstname cannot be Jim, sorry Jim']}]
```
poteto: Nitpick: prefer having spaces within `{}`, e.g. `[{ key: ... }]`

poteto: I'm not sure this addition helps with understanding how merge works, feels like it could be confusing

README.md
```diff
@@ -380,20 +380,24 @@ changeset.rollback(); // returns changeset
...
 changesetB.set('firstName', 'Jimmy');
+changesetB.get('errors') // []
 changesetB.set('lastName', 'Fallon');
 let changesetC = changesetA.merge(changesetB);
 changesetC.execute();
 user.get('firstName'); // "Jimmy"
 user.get('lastName'); // "Fallon"
 ```
 
+Changes are merged using similar heuristics to [`Object.assign`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign), such that `B` will replace `A` in `changesetA.merge(changesetB)`. Neither changeset is actually mutated, instead a clean changeset is created from `A` and `B` which contains `errors` from `A` and `B` where `errors` on `A` do not include valid keys on `B`, and valid keys on `A` do not include errors on `B`.
```
poteto: The last sentence is confusingly worded

ivanvanderbyl: Yeah... I need to eat something, will fix later.

addon/index.js
```diff
@@ -211,18 +212,28 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
...
       let changesA = get(this, CHANGES);
       let changesB = get(changeset, CHANGES);
-      let mergedChanges = pureAssign(changesA, changesB);
+
+      let errorsA = get(this, ERRORS);
+      let errorsB = get(changeset, ERRORS);
+
       let newChangeset = new Changeset(content, get(this, VALIDATOR));
+
+      errorsA = objectWithout(keys(changesB), errorsA);
```
poteto: Can we create a new variable?

tests/unit/utils/object-without-test.js
```diff
@@ -0,0 +1,15 @@
+import objectWithout from 'dummy/utils/object-without';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | object without');
+
+test('it exludes the given keys from all merged objects', function(assert) {
+  let objA = {name: 'Ivan'};
```
poteto: Minor formatting: `{ name: ... }`, etc for the rest of the objects in this file

README.md
ivanvanderbyl: Cleaned that up

ivanvanderbyl:
Ready to roll

README.md
```diff
@@ -380,20 +380,24 @@ changeset.rollback(); // returns changeset
...
 changesetB.set('firstName', 'Jimmy');
+changesetB.get('errors') // []
 changesetB.set('lastName', 'Fallon');
 let changesetC = changesetA.merge(changesetB);
 changesetC.execute();
 user.get('firstName'); // "Jimmy"
 user.get('lastName'); // "Fallon"
 ```
 
+When merging, the changeset being merged in takes precedence. For example: if `B` contains errors, the errors from `B` will be merged, and the invalid attributes on `B` will not be merged from `A` even if they're valid on `A`. Likewise, if `A` contains errors on attributes which are valid on `B`, the value from `B` will be used and the errors for those keys will not be merged from `A`. This is to prevent the new changeset from containing `changes` and `errors` on the same keys, which is a guarantee of a normal changeset.
```
poteto: This is still confusing, in fact I think it would be better to just remove all the changes to the readme and only keep the change on L383. I feel like this is just implementation detail that the end user doesn't need to know about.

addon/index.js
```diff
@@ -211,18 +212,28 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
...
-      assert('Cannot merge invalid changesets', get(this, 'isValid') && get(changeset, 'isValid'));
 
       if (get(this, 'isPristine') && get(changeset, 'isPristine')) {
         return this;
       }
 
       let changesA = get(this, CHANGES);
       let changesB = get(changeset, CHANGES);
-      let mergedChanges = pureAssign(changesA, changesB);
+
```
poteto: No whitelines between the variable declarations please!

poteto:
Sorry about the delay, was out the whole day! Just a minor update to the readme and we're good to go

addon/index.js
ivanvanderbyl: Do you think you could add a `jcscrc` or `eslint` config with the rules you're after?

poteto: Yeah I will at some point

ivanvanderbyl:
@poteto :shipit: ?

poteto:
Yes, thanks for this!

PR #75, opened 2016-07-19
Update README.md

dustinfarris:
Account for old value when adding error in ember data example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,7 +505,7 @@ changeset.save()
   .then(() => { /* ... */ })
   .catch(() => {
     get(this, 'model.errors').forEach(({ attribute, message }) => {
-      changeset.addError(attribute, { validation: message });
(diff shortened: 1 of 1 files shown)


dustinfarris:
@poteto I wonder if the old value could be preserved if `value` is not given to addError

README.md
```diff
@@ -505,7 +505,8 @@ changeset.save()
   .then(() => { /* ... */ })
   .catch(() => {
     get(this, 'model.errors').forEach(({ attribute, message }) => {
-      changeset.addError(attribute, { validation: message });
+      let oldValue = get(changeset, attribute);
+      changeset.addError(attribute, { value: oldValue, validation: message });
```
poteto: Maybe just value: get(changeset, attribute)

dustinfarris:
Ready

PR #77, opened 2016-07-22
Add shortcut to addError method.

acburdine:
So I'm totally willing to have this closed if it's not something wanted; however, it is a change that made several things easier in the Ghost validations implementation.

Basically, this allows one to only pass in a string as the second argument to `addError`. If that is done, the value that exists in the changeset for the specified key will be used as the key for the error.


acburdine:
Build just needs a restart 😄

PR #84, opened 2016-07-28
[DOC] Removed duplicate ####Rollback documentation

jeffreybiles:
There were two of them, so I removed the second.  If there was a reason for having it two times, then please ignore.


jeffreybiles:
Closing in favor of #87, which has the desired fix

PR #87, opened 2016-08-03
[DOC] Removed duplicate ####Rollback documentation

jeffreybiles:
This updates #84, but removes the first one instead of the second (as shown in ToC)


poteto:
Thanks!

PR #99, opened 2016-08-25
Always persist changes on #save

bugant:
Even if the content object does not implement the save function, persists the changes on #save and returns a Promise.


bugant:
Note that tests are OK, there is  an error on connecting to bower but I don't know how to force circle-ci to re-run tests.

PR #109, opened 2016-09-06
Mark changeset's pending/fulfilled/valid states

### LINKED ISSUE #102: Async validator question
Hi,

Is there a way to check if a changeset has changed since it was last validated?

Ideally setting any prop: `changeset.set('prop', 1)` will turn `changeset.get('isValid')` false, until all validators (async here is the issue) resolve.

This would allow us to disable UI until we know for certain that everything is resolved and valid.

Currently, if there is a slow async validator, the changeset will be in `isValid` true, but in reality, it is still unresolved, and therefor, the user may proceed to submit and then the world falls apart.

Let me know what you think.

Thanks

poteto:
... when async validator(s) are running

Closes #102

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,8 @@ On rollback, all changes are dropped and the underlying Object is left untouched
   + [`isInvalid`](#isinvalid)
   + [`isPristine`](#ispristine)
   + [`isDirty`](#isdirty)
+  + [`isPending`](#ispending)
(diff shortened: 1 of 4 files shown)


poteto:
I think this PR needs a bit more work to get over the line. I'm considering `ember-concurrency` as a dependency in order to get the async logic right.

PR #128, opened 2016-10-24
Add `pushErrors` method

poteto:


fotinakis:
Awesome! Ideally this would be named `pushError` instead of `pushErrors` because you're pushing a singular error into an array (like [pushObject](http://emberjs.com/api/classes/Ember.MutableArray.html#method_pushObject)). Not sure if it's too late to change.

poteto:
I intentionally named it `pushErrors` because you can push multiple.

```js
changeset.pushErrors('firstName', 'foo', 'bar', 'baz');
```

fotinakis:
Ah! Didn't read the implementation close enough, thanks.

PR #130, opened 2016-11-05
Changeset helper handles promise as model

### LINKED ISSUE #127: Building a changeset from 'belongs-to' association (question)
Hi, 

I'd like to know if this is posible. I want to create a changeset from a belongs-to association. I have a User model associated to one Location. So, in my template I wrote this:

{{location-form changeset=(changeset model.location)}}

but that doesn't build the changeset properly.

Thanks.

poteto:
Closes #127

### DIFF
--- a/addon/helpers/changeset.js
+++ b/addon/helpers/changeset.js
@@ -1,12 +1,17 @@
 import Ember from 'ember';
 import Changeset from 'ember-changeset';
 import isChangeset from 'ember-changeset/utils/is-changeset';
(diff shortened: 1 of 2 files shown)


mike183:
Would you considering supporting promises when creating a changeset using JS? 

E.g.

```js
let changeset = new Changeset(somePromise, validatorFn);
```

poteto:
@mike183 I would expect the user to resolve the promise and then create the changeset

PR #131, opened 2016-11-07
Add support for executing changes on hasMany relations

gmarkv10:
I commented earlier in one of the issue threads on ember-changeset-validations. This fix should solve the issue I was having. Thanks for your consideration.


sandstrom:
Found this addon and it's excellent! 💯  We're using something home-baked that is fairly similar (but this one is better).

Also curious about relationship changes. What's the status here? @poteto @gmarkv10

sandstrom:
Friendly ping @Dhaulagiri and @snewcomer 😄

sandstrom:
@snewcomer Certainly! 👍

snewcomer:
I verified that adding and removing has many relationships on master works.  Lmk if anybody has any other issues!

Issue #133, opened 2016-11-18
[doc] Best-practise to use with complex objects


marxsk:
I'm using ember-changeset with more complex objects. Let's assume:

    this.properties = Ember.A();
    this.properties.pushObject({name: 'name', level: 'advanced', type: 'text', title: 'My Name', value: '123'});
    this.properties.pushObject({name: 'city', type: 'select', title: 'City', value: 'Brno', options: ['Brno', 'Praha', 'New York']});
    this.properties.pushObject({name: 'author', level: 'advanced', type: 'text', title: 'Author', value: null, defaultValue: 'Hugo Hugo'});
 
what differs from expected

    this.properties = { name: 'My Name', city: 'Brno', author: null }

My question is what is the best practise to use more complex objects? 

Currently, my solution is to use computed properties to generate expected format from my complex one. It works but I believe there might be a better option.

Mithrilhall:
@marxsk Could you post what you're doing with computed properties? I'm running into the same problem.

marxsk:
@Mithrilhall Sure. I don't have a final commit on github yet but I had to create a component that encapsulate form generation to be able to use computed property.

    simpleProperties: Ember.computed('properties.@each.value', {
       get() {
       let ret = {};
       this.get('properties').forEach(function(x) {
          ret[x.get('name')] = x.get('value');
       });
       return ret;
    }}),

The explicit get() is there because I was thinking about setting computed property. Later I have decided that it would not help me that much. 

In the controller, I have following action:

    submitProperties: function(properties, changeset) {
      // update values locally
      for (let attrName in changeset.get('change')) {
        properties.forEach(function(item, index) {
          if (item.get('name') === attrName) {
            item.set('value', changeset.get(attrName));
          }
        });
      }

      // save changes to remote server
      this.store.pushClusterProperties(changeset);
    }

Mithrilhall:
@poteto Is flattening the object the recommended way to deal with nested objects?

From:

```{
  "firstName": "William",
  "lastName": "Wallace",
  "address": {
    "country": "United States of America",
    "state": {
      "name": "Alaska",
      "abbreviation": "AK",
      "town": {
        "name": "Barrow",
        "coordinates": {
          "latitude": "71.29056",
          "longitude": "-156.788666"
        }
      }
    }
  }
}
```

To:

```
{
  "firstName": "William",
  "lastName": "Wallace",
  "address.country": "United States of America",
  "address.state.name": "Alaska",
[...]

PR #141, opened 2016-12-09
Feature nest support

ppegusii:
@poteto I saw that you were working on support for nested objects and wanted to show you my WIP on the feature. It returns proxy objects when `_valueFor(key)` would return an object or array. The proxy objects returns get/set control back to changeset. Arrays of primitives don't yet work because Ember.ArrayProxy does not provide a proxy setter for an index AFAIK. I plan to try other proxies to address the issue.


remithomas:
@ppegusii what about the typeof _instance_ ?

ppegusii:
@remithomas good point. I'll add it.

wmadden:
To completely support nesting of arrays and objects of arbitrary depth, `ember-changeset` would need to recurse through the properties of the object it wraps, as well as every value of any arrays. It might be better to exclude nesting this from the responsibilities of the library, since it adds a lot of complexity.

You could just throw an error if the consumer attempts to `set()` a property whose value is an object or array, so at least consumers will have a hint on how to handle it themselves (and not end up here or in #133).

kevlarr:
Why not set an object/array property using `JSON.parse(JSON.stringify(...))`.. ?

Properties like this generally store data-only objects, not objects that contain functions.

nucleartide:
Hey folks, since https://github.com/poteto/ember-changeset/pull/211 was merged, I'm going to go ahead and close this. :( Feel free to open a new PR if you have any additions.

PR #143, opened 2016-12-27
Support complex objects, don't mark changes from validate if not changed

### LINKED ISSUE #132: `validate` should not set changes
Currently, when calling `validate` on a changeset, all changes will also be set on the changeset. `validate` should only set errors.

bgentry:
This is my attempt at fixing #132. I'm not sure if the description there was accurate (`validate` only setting `errors` and never setting `changes`). I tried doing that, but a ton of tests broke. I settled on an implementation that only touches `changes` for keys that have actually changed.

Fixes #132. Hopefully!

*Update*: Also includes the change from #144.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,7 +1,6 @@
 import Ember from 'ember';
 import objectToArray from 'ember-changeset/utils/computed/object-to-array';
 import isEmptyObject from 'ember-changeset/utils/computed/is-empty-object';
-import objectEqual from 'ember-changeset/utils/computed/object-equal';
 import isPromise from 'ember-changeset/utils/is-promise';
 import isObject from 'ember-changeset/utils/is-object';
 import pureAssign from 'ember-changeset/utils/assign';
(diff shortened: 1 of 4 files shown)


addon/index.js
```diff
@@ -451,7 +450,12 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
 
       if (validation === true || isSingleValidationArray) {
         this._deleteKey(ERRORS, key);
-        set(changes, key, value);
+
+        if (oldValue !== value) {
```
bgentry: I'm guessing this is not a good enough way to compare equality. It's possible that both of these are objects (as I get using ember-ted-select), so this would not properly detect equality there.

bgentry:
I edited this so that it uses `Ember.isEqual` to determine equality. That was easier to test with the changes in #144, so I rebased this commit on top of that one and this PR now includes both.

Let me know if you want me to try to split this out somehow.

poteto:
Sorry about the delay, I'll review this shortly!

bgentry:
No worries! Let me know what I can fix.

I've been using this in my app for a bit now with no real issues. Though #140 is related and would make nested object support more complete.

PR #144, opened 2016-12-27
support complex values

bgentry:
Don't flatten complex values into the changes object.

Prior to this, if my object looked like:

```js
{ name: 'foo', options: { awesomeMode: true } }
```

After validation, `changeset.get('changes')[0]` would be:

```js
{ key: 'options', awesomeMode: true }
```

With this change, it instead returns:

```js
{ key: 'options', value: { awesomeMode: true }}
```


bgentry:
included in #143 so I'm closing this one.

PR #148, opened 2017-02-12
Add option to skip validations when setting values

alexgaribay:
This adds the option to skip validations when setting values to the underlying object. This option would only be available when creating the changeset programmatically by using a forth parameter. The user would be responsible for manually invoking `validate` on their changeset.

In combination with ember-changeset-validations, it would looks like this:

```
  ...
  model() {
    return new Changeset(
       someObject,
       lookupValidator(SomeValidator),
       SomeValidator,
       { skipValidationsOnSet: true }
    );
  }
  ...
```

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -29,10 +29,13 @@ const CONTENT = '_content';
 const CHANGES = '_changes';
 const ERRORS = '_errors';
 const VALIDATOR = '_validator';
+const OPTIONS = '_options';
 
 function defaultValidatorFn() {
   return true;
 }
+
+const defaultOptions = { skipValidationsOnSet: false };
 
 /**
  * Creates new changesets.
@@ -40,9 +43,10 @@ function defaultValidatorFn() {
  * @param  {Object} obj
  * @param  {Function} validateFn
  * @param  {Object} validationMap
+ * @param  {Object}  options
  * @return {Ember.Object}
  */
-export function changeset(obj, validateFn = defaultValidatorFn, validationMap = {}) {
+export function changeset(obj, validateFn = defaultValidatorFn, validationMap = {}, options = {}) {
   assert('Underlying object for changeset is missing', isPresent(obj));
 
   return EmberObject.extend({
@@ -71,6 +75,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
       this[CHANGES] = {};
       this[ERRORS] = {};
       this[VALIDATOR] = validateFn;
+      this[OPTIONS] = Object.assign({}, defaultOptions, options);
     },
 
     /**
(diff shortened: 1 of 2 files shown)


addon/index.js
```diff
@@ -71,6 +75,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
       this[CHANGES] = {};
       this[ERRORS] = {};
       this[VALIDATOR] = validateFn;
+      this[OPTIONS] = Object.assign({}, defaultOptions, options);
```
poteto: Could you use `pureAssign` here please?

addon/index.js
```diff
@@ -93,7 +98,15 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
...
      */
     setUnknownProperty(key, value) {
-      return this._validateAndSet(key, value);
+      let changesetOptions = get(this, OPTIONS);
+      let skipValidationsOnSet = get(changesetOptions, 'skipValidationsOnSet');
+      
+      if (skipValidationsOnSet) {
+        return this._setProperty(true, { key, value });
+      }
+      else {
```
poteto: This `else` block is unnecessary, just `return this._validateAndSet(key, value);` is sufficient

addon/index.js
```diff
@@ -29,20 +29,24 @@ const CONTENT = '_content';
 const CHANGES = '_changes';
 const ERRORS = '_errors';
 const VALIDATOR = '_validator';
+const OPTIONS = '_options';
 
 function defaultValidatorFn() {
   return true;
 }
 
+const defaultOptions = { skipValidationsOnSet: false };
```
poteto: Can we rename the option to just `skipValidate`?

alexgaribay:
I've updated the code to reflect your requests. Let me know if you want anything else changed!

README.md
```diff
@@ -114,6 +114,16 @@ In the above example, when the input changes, only the changeset's internal valu
 
 On rollback, all changes are dropped and the underlying Object is left untouched.
 
+## Disabling Automatic Validation
+
+The default behavior of `Changeset` is to automatically validate a field when it is set. Automatic validation can be disabled by passing `skipValidate` as on option only when creating a changeset programmatically.
```
poteto: This actually reminds me, we should let the option be passed into the helper as well:

```js
export function changeset([obj, validations], { skipValidate }) {
  //
}
```

alexgaribay:
Docs and helper have been updated to reflect the new option.

alexgaribay:
It looks like the tests failed for a reason unrelated to the code changes?

alexgaribay:
Is there anything else that you would want changed?

PR #158, opened 2017-03-17
Key detection that works for Ember Objects

demersus:
I'm not sure if this is the direction you were wanting to go, but this fixes the problem for ember objects.

See: https://github.com/DockYard/ember-changeset/issues/153

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -465,7 +465,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
 
         if (!isEqual(oldValue, value)) {
           set(changes, key, value);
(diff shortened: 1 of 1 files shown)


poteto:
@gordonbisnor can you give this a try and let me know if it fixes the issue for you?

gordonbisnor:
@poteto ha yes I had already done pull request #151 with the same change, so in my opinion this works.

poteto:
Ah my bad, I completely missed that. Closing this in favor of #151.

gordonbisnor:
Ha no worries thanks for the update!

PR #169, opened 2017-04-10
Remove changes on set when model is Ember.ObjectProxy

HeroicEric:
`key in obj` does not correctly detect properties that are defined on the underlying object of an `Ember.ObjectProxy`. Since this conditional is for deciding whether or not to delete a key in `changes`, I think it makes sense to check the `changes` object for its existence anyway.

Here's a twiddle showing the problem https://ember-twiddle.com/bac73fd0fd7cae267621638b9621a7bb?openFiles=controllers.application.js%2C


urbany:
@poteto or @HeroicEric would you mind making a new release with this change please?

PR #170, opened 2017-04-15
Check for changed key on changes, not on the parent changeset obj

kbbpartners:
I noticed that toggling a boolean value on a changeset would make a single change, but subsequent attempts to toggle the boolean were ignored. Digging into the `_setProperty` function, it looks like the `else if (key in obj)` is never triggered because `obj` is the whole parent changeset, in which the key is buried under `_changes` . By checking `get(this, CHANGES)` instead, we found the expected behavior—_i.e._, removing the change when old and new values match.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -495,7 +495,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
 
         if (!isEqual(oldValue, value)) {
           set(changes, key, value);
-        } else if (key in obj) {
+        } else if (key in changes) {
           delete changes[key];
         }
         this.notifyPropertyChange(CHANGES);


kbbpartners:
Interesting—it looks like this arose due to the fact that we were using a nested component that was making a changeset of a changeset, hence `obj` was itself a changeset.

Nevertheless, this change seems to support both use cases: (a) where the change is applied to a changeset based on a model, and (b) where the changeset is based on another changeset. Let me know if we've overlooked anything.

HeroicEric:
I think this is the same issue I ran into here https://github.com/DockYard/ember-changeset/pull/169

kbbpartners:
Yes, it's the same issue. Your PR has tests, which makes it the clear winner.

PR #181, opened 2017-05-11
Remove back-to-top links

sandstrom:
This is a great addon! 🌞 

May I kindly suggest you remove the back-to-top links, they make the readme harder to read (unnecessary text and emojis).

Something like `cmd` + `up-arrow` (macOS) is available in all browsers. Also, README's are generally read from top to bottom, or scrolled through.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,8 +181,6 @@ You can use this property to locate a single error:
 {{/if}}
 ```
 
-**[⬆️ back to top](#api)**
-
 #### `change`
 
 Returns the change object.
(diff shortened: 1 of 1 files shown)


nucleartide:
Hey @sandstrom, thanks for the PR!

You make a good point, but I think the intent of the back-to-top links is to provide a jumping point to the table of contents (as an aside, this is a neat feature for keyboard nav). I'd also like to avoid removing @poteto's stylings.

Going to close this, sorry! :(

PR #186, opened 2017-06-10
[Bug report] Added a failing test case for nested objects

shijistar:
We have some big forms that editing large objects which is composition of several smaller objects. I find it's hard to write validations directly, actually I don't know how to write them.  I'm inspired by [this post](https://github.com/DockYard/ember-changeset-validations/issues/110) and tried the same thing, but unfortunately failed. So I write a test case, hope it's going to be fixed or supported. Thanks.


shijistar:
Sorry, just noticed this work is in progress. I'm closing it.

PR #190, opened 2017-06-28
Fix: pass oldValue to _setProperty if `skipValidate` is `true`

jelhan:

PR #191, opened 2017-06-28
Fix: changeset.set('prop', undefined); does not update prop

jelhan:
`changeset.set('prop', undefined)` does not change value.

Please not that #190 is needed to fix this if `skipValidate` is `true`. Since without #190 `oldValue` is always `undefined`, condition `isEqual(oldValue, value)` will always be `true` if `value` is `undefined` and therefore change will not be executed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "homepage": "https://github.com/poteto/ember-changeset",
   "dependencies": {
     "ember-cli-babel": "^6.8.2",
-    "ember-deep-set": "^0.1.3"
+    "ember-deep-set": "^0.1.4"
   },
   "devDependencies": {
     "broccoli-asset-rev": "^2.7.0",
--- a/tests/unit/changeset-test.js
+++ b/tests/unit/changeset-test.js
@@ -94,6 +94,18 @@ test('#change returns the changes object', function(assert) {
   assert.deepEqual(get(dummyChangeset, 'change'), expectedResult, 'should return changes object');
 });
 
+test('#change supports `undefined`', function(assert) {
+  let model = { name: 'a' };
+  let dummyChangeset = new Changeset(model);
+  let expectedResult = { name: undefined };
+  dummyChangeset.set('name', undefined);
+
+  assert.deepEqual(
+    get(dummyChangeset, 'change'), expectedResult,
+    'property changed to `undefined` should be included in change object'
+  );
+});
+
 /**
  * #errors
  */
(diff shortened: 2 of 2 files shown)


addon/index.js
```diff
@@ -494,7 +494,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
         this._deleteKey(ERRORS, key);
 
         if (!isEqual(oldValue, value)) {
-          set(changes, key, value);
+          changes[key] = value;
```
jelhan: I'm not quite sure why `Ember.set` fails to update value. I think changing property directly might be okay since we call `this.notifyPropertyChange(CHANGES);` later.

BorigTheDwarf:
+1

snewcomer:
@jelhan Is this still a problem for ya?

jelhan:
Don't have the time today. I think we should check if this scenario is already tested in master. If not, we could apply the test in this PR. If it fails, we might use the fix provided by this PR. Will try to do it over the weekend.

jelhan:
Current code base is using provided patch. After a lot of refactoring the code line now lives at: https://github.com/poteto/ember-changeset/blob/master/addon/utils/set-nested-property.js#L41

After setting a property to `undefined`, `changes` contains the changed property. So original issue is fixed. But another one is added. `change` object does not contain a changed property with a value of `undefined`.

Simple reproduction:

```
changeset = new Changeset({ prop: 'foo' });
changeset.set('prop', undefined);
changeset.get('changes'); // [{ key: 'prop', value: undefined }]
changeset.get('change'); // {}
```

This PR now provides additional tests to cover both. Test for `change` property is failing.

jelhan:
Reported the issue regarding `change` upstream: https://github.com/poteto/ember-deep-set/pull/14

algodave:
I have an `address` model with a `belongsTo('state')` association. I create a changeset from it. In my edit address form the following happens:

*  when user selects a different `state`, it is correctly set in the changeset
*  when user selects a different `country`, the `state` in the changeset should be unset; I tried setting it to either `undefined` or `null`, but none of them worked: the old `state` model stays there in the changeset!

In both cases, I'm using `ember-changeset`'s overridden `set` method rather than Ember's standard `set`. I'm using a50fc86cbfe8c15bca38f990cd908cb2d2ab6d8f in my project.

jelhan:
@algodave You could try using [this pull request](https://github.com/poteto/ember-deep-set/pull/14) for ember-deep-set. You could do so by [selective dependency resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) feature of yarn. The pull request should fix this issue. Please report your findings here.

algodave:
Thanks @jelhan will do!

snewcomer:
@jelhan Looks like a minor merge conflict here. ember-deep-set 0.1.4 was published fyi!

jelhan:
Fixed merge conflicts, which were introduced by my own PR. :laughing: Also updated ember-deep-set@0.1.4. Don't get confused by travis passing even without that update. It uses `yarn install --no-lockfile`.

PR #193, opened 2017-07-20
Support non-promise return values from proxied save()

bendemboski:
There's no need to require the proxied object's save() method to return a promise rather than a simple value, and saving a changeset to some data source is not necessarily an asynchronous operation, so this commit adds support for non-promise return values from `save`.


bendemboski:
@poteto any chance of getting this released sometime soon?

PR #194, opened 2017-07-27
Updates to nested change set support.

webark:
Looking past the odd merge issues (which I think in reality are all solved even though it looks like a little bit of master made it's way in) here is an update to address the concerns.

### DIFF
--- a/addon/-private/relay.js
+++ b/addon/-private/relay.js
@@ -9,17 +9,26 @@ export default EmberObject.extend({
   changeset: null,
   key: null,
 
+  init() {
+    this._changedKeys = [];
+  },
+
   unknownProperty(key) {
     return this.changeset._valueFor(`${get(this, 'key')}.${key}`);
   },
 
   setUnknownProperty(key, value) {
-    const selfKey = get(this, 'key');
-    this.changeset._validateAndSet(`${selfKey}.${key}`, value);
-    if (!/\./.test(selfKey)) {
+    this._changedKeys.push(key);
+    this.changeset._validateAndSet(`${get(this, 'key')}.${key}`, value);
+    this.notifyPropertyChange(key);
+    return value;
+  },
+
+  rollback() {
+    for (let key of this._changedKeys) {
       this.notifyPropertyChange(key);
     }
-    return value;
+    this._changedKeys = [];
   },
 
   destroy() {
--- a/addon/-private/relay.js
+++ b/addon/-private/relay.js
@@ -10,20 +10,16 @@ export default EmberObject.extend({
   key: null,
 
   unknownProperty(key) {
-    return this.getValue(key);
+    return this.changeset._valueFor(`${get(this, 'key')}.${key}`);
   },
 
   setUnknownProperty(key, value) {
-    this.setValue(key, value);
+    const selfKey = get(this, 'key');
+    this.changeset._validateAndSet(`${selfKey}.${key}`, value);
+    if (!/\./.test(selfKey)) {
+      this.notifyPropertyChange(key);
+    }
     return value;
-  },
-
-  getValue(key) {
-    return this.changeset.valueFor(`${get(this, 'key')}.${key}`);
-  },
-
-  setValue(key, value) {
-    return this.changeset.validateAndSet(`${get(this, 'key')}.${key}`, value);
   },
 
   destroy() {
--- a/addon/index.js
+++ b/addon/index.js
@@ -9,6 +9,7 @@ import objectWithout from 'ember-changeset/utils/object-without';
 import includes from 'ember-changeset/utils/includes';
 import take from 'ember-changeset/utils/take';
 import isChangeset, { CHANGESET } from 'ember-changeset/utils/is-changeset';
+import deepSet from 'ember-deep-set';
 
 const {
   Object: EmberObject,
@@ -96,7 +97,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
      * @return {Any}
      */
     unknownProperty(key) {
-      return this.valueFor(key);
+      return this._valueFor(key);
     },
 
     /**
@@ -115,7 +116,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
         return this._setProperty(true, { key, value });
       }
 
-      return this.validateAndSet(key, value);
+      return this._validateAndSet(key, value);
     },
 
     /**
@@ -127,16 +128,6 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
     toString() {
       let normalisedContent = pureAssign(get(this, CONTENT), {});
       return `changeset:${normalisedContent.toString()}`;
-    },
-
-    /**
-     * Teardown relays from cache.
-     *
-     * @public
-     * @return {Void}
-     */
-    willDestroy() {
-      // TODO destroy all relays
     },
 
     /**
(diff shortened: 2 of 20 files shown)


webark:
ok, @poteto @willrax @nkgm fixed the merge issues. Any feedback you might have would be greatly appreciated!!

webark:
Would be nice to have some more support around distorting, merging, rolling back, etc. though.

webark:
@nkgm added that rollback child nested test in there, and added the fix for it.

webark:
I'm still not 100% in love with all the solutions, and at times feels like I'm applying bandages rather than fixing root causes.

webark:
@poteto did you get a chance to review this? Anyone else you know of that I could ping to review it?

webark:
@poteto @bcardarella @martndemus @willrax Just trying to reach out to someone to see who could look at this.

rwwagner90:
Would love to get this reviewed and merged! Perhaps @webark could be made a collaborator, so he could help review and merge things? Seems he has put quite a bit of work into this.

bcardarella:
I'll defer to @poteto on this change.

webark:
@bcardarella ya, no worries. Was just going through and finding anyone that had made a commit!!

nrcjbarry:
@webark I just pulled this fork into my project, but when I create a changeset, I get this error:

"Could not find module `ember-deep-set` imported from `ember-changeset/index`"

nkgm:
@nrcjbarry did you `yarn install`?

nrcjbarry:
I removed my `node_modules` directory and `package-lock.json` file, then did an `npm install`.

nkgm:
If it fails to install the transitive dependency, you can try installing manually with `ember install ember-deep-set`.

nrcjbarry:
@nkgm yeah that works, I just want to bring that to light because it *should* be installed.

webark:
@nrcjbarry :facepalm: i put deepest in dev dependencies. I can change that. Sorry!

nrcjbarry:
@webark, using this PR, the rollback for my nested model is working now. Only issue I’m seeing is when I actually try to save the changeset, I get this error:

`RangeError: Maximum call stack size exceeded`

My data structure looks like this:

```
{
  name: 'Bob',
  contact: {
    emails: [ 'bob@email.com', 'the_bob@email.com' ]
  }
}
```

I wasn't getting this error when using the latest `ember-changeset`. Any thoughts?

webark:
@nrcjbarry i have an idea. I'll add a test with that structure and update the PR

webark:
@nrcjbarry I just added that test, and it looks like it's working as expected. Can you take a look at https://github.com/DockYard/ember-changeset/pull/194/commits/e74c7ba4780e62767db98e757dca810777a0a6e8 and let me know if you are using it differently?

webark:
also just moved the deepset to a dep rather than a dev dep. Maybe try uninstalling ember-deep-set and ember-changeset and reinstalling from the fork.

webark:
hmm. Getting an odd error that isn't part of the work. Will try again.

nrcjbarry:
@webark that test looks like it would cover my need. I can give this a try once you fix the error you just ran into.

webark:
ya. Still getting some weird errors that are not matching up. Passing some versions, but others get a `Unexpected token u in JSON at position 0` and ```=== Scenario: ember-lts-2.4 ====================================================

Cannot find module 'is-absolute'
Error: Cannot find module 'is-absolute'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/ubuntu/ember-changeset/node_modules/npm/node_modules/which/which.js:12:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)

Result: false
---
```
so like 1.13 passes, 2.4 fails cause of the `is-absolute` 2.8 passes, 2.14 fails due to `is-absolute`, beta fails due to the `Unexpected token u in JSON at position 0`, and then canary fails again due to the  `is-absolute`.
Which none of these fails when I run them locally.

**Comment by webark** [comment]
temporarily upgrading ember-cli to see if it fixes the odd circle-cli issues.

**Comment by webark** [comment]
sorry for the commit spam. Just trying to situate out why circle ci is failing.

**Comment by nrcjbarry** [comment]
@webark what changes were made between when it was working and when it started barfing?

**Comment by webark** [comment]
just added more tests. Just trying to figure out what's going on. https://travis-ci.org/webark/ember-changeset Travis is passing. So it must be something with the circle ci config.

**Comment by webark** [comment]
ya. Every other scenario is failing with that `Cannot find module 'is-absolute'` error. So it makes me think that something isn't getting cleaned up properly.

**Comment by webark** [comment]
@poteto any more thoughts on merging this into the `feature/support-nested-keys` branch? Then we can wrap up anything over in #140 ? I can pull out the ember-cli updates if you'd prefer. Was just trying to figure out why circle ci was failing but travis was passing.

**Comment by webark** [comment]
ping. Still, would love to get some traction on this!!

**Comment by rwwagner90** [comment]
+1

File: addon/-private/relay.js [hunk 1]
```diff
@@ -9,21 +9,26 @@ export default EmberObject.extend({
   changeset: null,
   key: null,
 
+  init() {
+    this._changedKeys = [];
```
webark: This is just a list so that they can be properly notified when a change has occurred. And how can two relays modify the same end property? This is only storing the key of the object.

addon/index.js
```diff
@@ -184,9 +188,25 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
       if (get(this, 'isValid') && get(this, 'isDirty')) {
         let content = get(this, CONTENT);
         let changes = get(this, CHANGES);
+
```
webark: this was to be able to use "setProperties" due to the call of "setProperties" propey triggers an objects internal "dirty" state.. But, if there is no "deepSetProperties", so i have to remember all the properties that weren't set, and reset them after.

addon/index.js
```diff
@@ -538,20 +558,13 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
...
 
         if (!isEqual(oldValue, value)) {
-          this._recursivelySet(key, value, changes);
+          deepSet(changes, key, value);
         } else if (key in changes) {
           delete changes[key];
         }
-
         this.notifyPropertyChange(CHANGES);
         this.notifyPropertyChange(root);
```
webark: i used to have it where both the root and they key got triggered, but triggering both is redundant, and any child keys get triggered from the relay.

webark:
and i always appreciate feedback!

addon/-private/relay.js
nkgm: You're right, I can see now that such condition is not likely to arise. A couple of points on `rollback`:
* While I definitely see and have use for it, is it something we want on relays? In the same vein we could add `snapshot`, `validate` etc. Thoughts?
* Couldn't per-relay rollback be implemented on the changeset like `this.changeset._rollbackRelay(this.key)`? That could also be the way for those other methods.

addon/index.js
nkgm: But out of an `N` deep object graph, root will always be level `1`. Shouldn't it be notifying level `N-1`?

ssured:
@webark: I just played with your PR and it seems `firefox_config.js` is missing in the repo. Maybe that's part of the problem with circleci?

webark:
@ssured is thata a circle ci thing? wasn't a recent removal as far as I can tell.

webark:
and @poteto any chance this could get merged in to the `feature/support-nested-keys` so we can start finishing this up?

catz:
Eagerly want to try this out

PR #205, opened 2017-09-28
Fix for isValid property not updating after a change

sandersoeteman:
Error changes are notified before the _errors object is changed. This caused the isValid property to lag


snewcomer:
@sandersoeteman I'm sorry it's been so long, but could you verify this may still be a bug?  Otherwise, we can close it as some changes have been made since this PR.

snewcomer:
Will close unless reproduced...

PR #207, opened 2017-10-19
Bump Ember CLI, Ember

mixonic:
The narrow goal here is to bump `ember-cli-babel` to 6.6+, which will permit consumers to drop their dependency on `ember-cli-shims`.

However it seems good simply to bump the Ember CLI version in general, so thats what I've done here. The non-dependency changes are all from `ember init` (looks like a few bugs were caught by that as well).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,10 @@
 /coverage/*
 /libpeerconnection.log
 npm-debug.log*
+yarn-error.log
 testem.log
+
+# ember-try
+.node_modules.ember-try/
+bower.json.ember-try
+package.json.ember-try
--- a/circle.yml
+++ b/circle.yml
@@ -3,9 +3,9 @@ machine:
     version: "6"
 dependencies:
   pre:
-    - npm install -g bower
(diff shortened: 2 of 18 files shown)


mixonic:
Hm running `yarn test` locally I see https://github.com/lytics/ember-cli-release has a dependency on `npm@~3.5.2` https://github.com/lytics/ember-cli-release/blob/master/package.json#L30 which seems to be failing on some dependency installation stuff ember-try fires off.

screen shot 2017-10-19 at 9 45 52 am

However this is *not* the reason Circle CI is failing. Circle CI fails because of:



Which seems unrelated.

mixonic:
I believe the issue was https://github.com/ember-cli/ember-try/issues/141 upstream in ember-try. Adding bower as a development dependency is the temporary workaround I've added here.

nucleartide:
@mixonic This looks good to me, thanks! It seems like we can also remove Bower now that https://github.com/ember-cli/ember-try/issues/141 is resolved – I'll do that in a separate PR to expedite things.

PR #208, opened 2017-10-25
Creating a model on execute and stacked changesets

lukaiser:
I know this works a little bit against #140 but what do you guys think?

I added two things:
A) You can create a changeset without a model. The model then gets created when you execute the changset. A user can for example fill out a form without the need to create a ember data model beforehand and therefore you don't have to clean it up if he uses the back button or anything else.

`dummyModel = new NewEDModel("myModelName", Store.create);`
` let dummyChangeset = new Changeset(dummyModel);`
 `dummyChangeset.execute();`

On execute the Store.create function gets called with the "myModelName" string. Then the content value gets replaced with the new created model and then the changes get applied.
I am not happy with the class name yet but maybe somebody else has an idea.

B) Instead of setting a object you can set a subchangeset instead. This works for values and objects in arrays. When executing the changeset, it looks for subchangesets and if they are valid. If all are valid all the changesets get executed.

`let dummyChangeset = new Changeset(dummyModel);`
 ` let dummyChangeset2 = new Changeset(dummyModel2);`
  `let dummyChangeset3 = new Changeset(dummyModel3);`
  `let dummyChangeset4 = new Changeset(dummyModel4);`
  `dummyChangeset.set('a', 'a');`
  `dummyChangeset.set('b', dummyChangeset2);`
  `dummyChangeset.set('c', 'c');`
  `let arr = ['a', dummyChangeset3, 'c', dummyChangeset4, 'd'];`
  `dummyChangeset.set('d', arr);`
  `dummyChangeset.set('e', 'e');`
  `dummyChangeset.execute()`

Not as automatic as nested keys but it works well with ember data and relations as well as with validation. What do you guys think?

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -11,7 +11,7 @@ import isChangeset, { CHANGESET } from 'ember-changeset/utils/is-changeset';
 
 const {
   Object: EmberObject,
-  RSVP: { all, resolve },
+  RSVP: { all, allSettled, resolve },
   computed: { not, readOnly },
   A: emberArray,
   assert,
@@ -29,6 +29,8 @@ const CONTENT = '_content';
 const CHANGES = '_changes';
 const ERRORS = '_errors';
 const VALIDATOR = '_validator';
+const MODELNAME = 'modelName';
+const STORE = 'store';
 
 function defaultValidatorFn() {
   return true;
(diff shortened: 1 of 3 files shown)


nucleartide:
Hey @lukaiser, thanks for the PR! Some responses:

> A) You can create a changeset without a model.

I don't see much of a benefit here. It isn't that much boilerplate to construct new models and pass them to `new Changeset()`. We'd also like to use a changeset with non-ED models.

> B) Instead of setting a object you can set a subchangeset instead.

We have "child" changesets as part of @webark's PR, see: https://github.com/poteto/ember-changeset/blob/cae85fd49eef957e8b7908252860b6be56f3a04d/tests/unit/changeset-test.js#L922-L941

Sadly, I'm going to close this. :( Thanks for the effort, and feel free to open another issue if you have any more suggestions.

PR #226, opened 2017-11-29
pulled out method which recursively returns all keys of an object

### LINKED ISSUE #225: Since nested keys work, I'm getting an error in the _allKeys method: object.hasOwnProperty is not a function
After revving `ember-changeset` to `1.4.1` and `ember-changeset-validations` to `1.2.9` (from `1.2.0` and `1.2.4` respectively), with no other changes in my code, I'm getting an error with the `hasOwnProperty` call in the [`_allKeys`](https://github.com/poteto/ember-changeset/blob/cae85fd49eef957e8b7908252860b6be56f3a04d/addon/index.js#L474) method of `ember-changeset/addon/index.js` when trying to validate & save my model:
```
_allKeys(object, keys = [], scope = '') {
    for (let key in object) {
        if (!object.hasOwnProperty(key)) {
```

Debugging the error shows `object` to be the following:
```
{
    clearNext: ƒ (id),
    clearTimeout: ƒ (id),
    next: ƒ (fn),
    now: ƒ (),
    setTimeout: ƒ (fn, ms)
}
```

Searching around, the only similarly structured object I can find is in a library called [BackburnerJS](https://github.com/BackburnerJS/backburner.js/blob/master/lib/index.ts#L68), and this ties in with what I'm seeing the value of `scope` set to (`"response.changeset._content.store._backburner._platform."`).

What I've no idea about is how `object` — which returns `'object'` for a call to `typeof(object)` — doesn't have the `hasOwnProperty` method from `Object.prototype`. Having said that, `object.constructor` shows `undefined`, so I guess that the object was created using some sort of dark arts.

I thought this might have been related to [Issue 220](https://github.com/poteto/ember-changeset/issues/220), but possibly not.

Would it cause a lot of problems to change `if (!object.hasOwnProperty(key)) {` to `if (!(key in object)) {`?

I see there are a few other uses of `.hasOwnProperty`, but am not sure as to whether they would need changing as well.

webark:
Now more independent and testable. Switched to using Object.keys.

fixes #225

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -4,7 +4,7 @@ import objectToArray from 'ember-changeset/utils/computed/object-to-array';
 import isEmptyObject from 'ember-changeset/utils/computed/is-empty-object';
 import isPromise from 'ember-changeset/utils/is-promise';
 import isObject from 'ember-changeset/utils/is-object';
-import recursiveKeys from 'ember-changeset/utils/recursive-keys';
+import leafKeys from 'ember-changeset/utils/leaf-keys';
 import pureAssign from 'ember-changeset/utils/assign';
 import objectWithout from 'ember-changeset/utils/object-without';
 import includes from 'ember-changeset/utils/includes';
@@ -190,7 +190,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
         let content = get(this, CONTENT);
         let changes = get(this, CHANGES);
 
-        let changedKeys = recursiveKeys(changes);
+        let changedKeys = leafKeys(changes);
         let resetProperties = {};
 
         changedKeys.forEach(function(key) {
@@ -202,7 +202,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
 
         setProperties(content, changes);
 
-        recursiveKeys(resetProperties).forEach(function(key) {
+        leafKeys(resetProperties).forEach(function(key) {
           if (!changedKeys.includes(key)) {
             deepSet(content, key, get(resetProperties, key));
           }
--- a/addon/index.js
+++ b/addon/index.js
@@ -4,6 +4,7 @@ import objectToArray from 'ember-changeset/utils/computed/object-to-array';
 import isEmptyObject from 'ember-changeset/utils/computed/is-empty-object';
 import isPromise from 'ember-changeset/utils/is-promise';
 import isObject from 'ember-changeset/utils/is-object';
+import recursiveKeys from 'ember-changeset/utils/recursive-keys';
 import pureAssign from 'ember-changeset/utils/assign';
 import objectWithout from 'ember-changeset/utils/object-without';
 import includes from 'ember-changeset/utils/includes';
@@ -189,7 +190,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
         let content = get(this, CONTENT);
         let changes = get(this, CHANGES);
 
-        let changedKeys = this._allKeys(changes);
+        let changedKeys = recursiveKeys(changes);
         let resetProperties = {};
 
         changedKeys.forEach(function(key) {
@@ -201,7 +202,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
 
         setProperties(content, changes);
 
-        this._allKeys(resetProperties).forEach(function(key) {
+        recursiveKeys(resetProperties).forEach(function(key) {
           if (!changedKeys.includes(key)) {
             deepSet(content, key, get(resetProperties, key));
           }
@@ -459,29 +460,6 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
       set(this, CHANGES, casted);
 
       return this;
-    },
-
-    /**
-     * Executes the changeset and saves the underlying content.
-     *
-     * @private
-     * @param  {Array} [keys=[]] Array of keys to push to
-     * @param  {String} [scope=''] A sring that can be prepended to the key.
-     * @return {Array}
-     */
-    _allKeys(object, keys = [], scope = '') {
-      for (let key in object) {
-        if (!object.hasOwnProperty(key)) {
-          continue;
-        }
-
-        if (isObject(object[key])) {
-          this._allKeys(object[key], keys, scope + key + '.');
-        } else {
-          keys.push(scope + key);
-        }
-      }
-      return keys;
     },
 
     /**
--- a/addon/utils/recursive-keys.js
+++ b/addon/utils/recursive-keys.js
@@ -37,7 +37,7 @@ const recursiveKeys = function(object = {}, scope = '') {
     const value = object[key];
     let newKeys = [scope + key];
 
-    if (!isArray(value) && isObject(value) && Object.keys(value).length) {
+    if (isObject(value) && !isArray(value) && Object.keys(value).length) {
       newKeys = recursiveKeys(value, scope + key + '.');
     }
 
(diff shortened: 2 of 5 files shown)


webark:
@nucleartide and/or @BillyRayPreachersSon Let me know if there are other test cases that you'd like to see.

nucleartide:
Nice! Let's also add a test for `Object.create(null)`, which doesn't have any prototype methods. 🙂

webark:
@nucleartide added in that test, and fixed the ember object test. Also switched it from a forEach to a reduce. Seems a lot simpler now. Because it was basically rewritten, I force pushed over that commit. Let me know if you have any other edgecases you'd like addressed.

addon/index.js
```diff
@@ -4,6 +4,7 @@ import objectToArray from 'ember-changeset/utils/computed/object-to-array';
 import isEmptyObject from 'ember-changeset/utils/computed/is-empty-object';
 import isPromise from 'ember-changeset/utils/is-promise';
 import isObject from 'ember-changeset/utils/is-object';
+import recursiveKeys from 'ember-changeset/utils/recursive-keys';
```
nucleartide: Let's change the name of this util to `leafKeys`, since this util excludes intermediary keys.

addon/index.js
```diff
@@ -201,7 +202,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
 
         setProperties(content, changes);
 
-        this._allKeys(resetProperties).forEach(function(key) {
+        recursiveKeys(resetProperties).forEach(function(key) {
           if (!changedKeys.includes(key)) {
```
nucleartide: I missed this, but let's wrap `changedKeys` with `Ember.A` here to ensure the `.includes` method exists.

nucleartide: (Unsure what the new Ember import would be here.)

addon/utils/recursive-keys.js
```diff
@@ -0,0 +1,48 @@
+import isObject from 'ember-changeset/utils/is-object';
+import { isArray } from '@ember/array';
+
+/**
+ * Recursivly loops over all sub object and
```
nucleartide: Let's add something about how this function only returns child-most (or leaf-most) keys. Example:

```js
/**
 * Given an object, return an array of the leaf-most keys
 * in nested string format.
 */
```

addon/utils/recursive-keys.js
```diff
@@ -0,0 +1,48 @@
...
+ * returns an array of keys in nesteted string format.
+ *
+ * eg. this object
```
nucleartide: Let's remove this example, I'd rather have people look at the tests.

addon/utils/recursive-keys.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * [
+ *   'item.innerItem.thing',
+ *   'item.innerItem.other',
+ *   'item.innerItem.nest.thing',
+ *   'item.innerItem.nest.other',
+ * ]
+ *
+ * @param  {Object} [object={}] Object whos keys to iterate over.
+ * @param  {String} [scope=''] A sring that can be prepended to the key.
```
nucleartide: Nitpick: change to `A string that can be prepended to the object's keys.`

addon/utils/recursive-keys.js
```diff
@@ -0,0 +1,48 @@
...
+ * into in array of keys
+ *
+ * [
+ *   'item.innerItem.thing',
+ *   'item.innerItem.other',
+ *   'item.innerItem.nest.thing',
+ *   'item.innerItem.nest.other',
+ * ]
+ *
+ * @param  {Object} [object={}] Object whos keys to iterate over.
```
nucleartide: Nitpick: change to `Object to inspect.`

addon/utils/recursive-keys.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param  {String} [scope=''] A sring that can be prepended to the key.
+ * @return {Array}
+ */
+const recursiveKeys = function(object = {}, scope = '') {
```
nucleartide: Let's make this `scope` param private. So:

```js
export default function leafKeys(object = {}) {
  return helper(object);
}

function helper(object = {}, scope = '') {
  // ...
}
```

tests/unit/utils/recursive-keys-test.js
```diff
@@ -0,0 +1,152 @@
+import recursiveKeys from 'ember-changeset/utils/recursive-keys';
+import { module, test } from 'qunit';
+
+import EmberObject from '@ember/object';
+
+module('Unit | Utility | recursive-keys');
+
+test('it returns an array with keys as nested strings', function(assert) {
```
nucleartide: Since we're repeating the same assertion, let's write these tests in table-driven form. Example:

```js
let testData = [
  {
    desc: 'it returns an array with keys as nested strings',
    input: {
      topValue: -1,
      item: {
        secondValue: 0,
        innerItem: {
        thing: 1,
        other: 2,
        nest: {
          thing: 3,
          other: 4
        }
      }
    },
    expectedResult: [
      'topValue',
      'item.secondValue',
      'item.innerItem.thing',
      'item.innerItem.other',
      'item.innerItem.nest.thing',
      'item.innerItem.nest.other',
    ]
  },

  // ...
];

testData.forEach(({ desc, input, expected }) => {
  test(desc, function(assert) {
    assert.deepEqual(leafKeys(input), expected, desc);
  });
});
```

tests/unit/utils/recursive-keys-test.js
```diff
@@ -0,0 +1,152 @@
...
+test('it treats null and empty objects as null values', function(assert) {
+  const nestedObject = {
+    topValue: -1,
+    item: {
+      secondValue: 0,
+      innerItem: {
+        thing: 1,
+        other: 2,
+        empty: {},
+        nest: Object.create(null),
```
nucleartide: Let's add a test case where `Object.create(null)` isn't at the "leaf" level. So for example:

```js
let o = Object.create(null);
o.foo = 'bar';
o.hello = 'world';
o.someObj = { blah: 'blah' };

nestedObject.item = o;
```

BillyRayPreachersSon:
Thanks both for the quick turnaround, this looks like it should fix my problem. Once this is released, I'll update my app and let you know how it goes.

webark:
@nucleartide I think I covered all of your comments. Let me know if you have any more.

nucleartide:
@webark Looks amazing! Thanks, going to merge this.

sfanetti:
I have a model that has a belongsTo relationship with another DS model.  The DS model - has a cyclical relationship within it because of an included service.  

This caused the stack to blow out with the previously implemented release.  Though I have not yet tested this - I suspect it will also have the same issue.  Any models that happen to internal circular relationships will get blown out with this. 

It would be good to have a schema of properties to set when executing the changeset instead of doing the recursive scan.  This limits the work done in the method and provides better control over what gets copied from the changeset to the saving one.  Do you have any suggestions about a strategy for approaching this within this library?

Maybe at construction - options can be passed in with a structure that outlines the relevant attributes that are mutable within the form so that nesting depth can be controlled?

nucleartide:
@sfanetti Please see my comment at https://github.com/poteto/ember-changeset/issues/239#issuecomment-350291579.

BillyRayPreachersSon:
I've just tested the latest code and I seem to have a slightly different stack issue than @sfanetti (based on @nucleartide's comment in https://github.com/poteto/ember-changeset/issues/239).

I've put the details into https://github.com/poteto/ember-changeset/issues/244.

PR #227, opened 2017-12-01
Fix error display for nested keys

### LINKED ISSUE #224: Errors for nested keys don't go away once shown
Trying to extend @poteto's original twiddle but there are still some issues. Need to take another look before publishing a patch release.

nucleartide:
Closes #224.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -549,7 +549,15 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
 
         let errors = get(this, ERRORS);
         if (errors['__ember_meta__'] && errors['__ember_meta__']['values']) {
-          delete errors['__ember_meta__']['values'][key];
+          let path = key.split('.');
+          if (path.length === 1) {
+            delete errors['__ember_meta__']['values'][key];
+          } else {
+            let branch = path.slice(0, -1).join('.');
+            let [leaf] = path.slice(-1);
+            let obj = get(errors, `__ember_meta__.values.${branch}`);
+            delete obj[leaf];
+          }
           set(this, ERRORS, errors);
         }
 
--- a/addon/index.js
+++ b/addon/index.js
@@ -681,18 +681,28 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
      * @param  {String} key
      * @return {Void}
      */
-    _deleteKey(objName, key) {
+    _deleteKey(objName, key = '') {
       let obj = get(this, objName);
 
       if (isNone(obj)) {
         return;
       }
 
-      if (obj.hasOwnProperty(key)) {
+      let keyPath = key.split('.');
+      let isNestedKey = keyPath.length > 1;
+
+      if (isNestedKey) {
+        let path = keyPath.slice(0, -1).join('.');
+        let leaf = keyPath.slice(-1)[0];
+        let branch = get(obj, path);
+        branch && delete branch[leaf];
+        return;
+      } else if (obj.hasOwnProperty(key)) {
         delete obj[key];
-        this.notifyPropertyChange(`${objName}.${key}`);
-        this.notifyPropertyChange(objName);
       }
+
+      this.notifyPropertyChange(`${objName}.${key}`);
+      this.notifyPropertyChange(objName);
     }
   });
 }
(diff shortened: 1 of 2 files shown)


nucleartide:
Not so sure about this Ember 1.13 compat stuff:

https://github.com/poteto/ember-changeset/blob/1ac6cf99f88945983765e6fb4a2f9cbe19938639/addon/index.js#L550-L554

Let's see if CI passes.

webark:
@nucleartide why are you using two inputs with the same id?

nucleartide:
@webark Whoops, that's a mistake.

addon/index.js
```diff
@@ -549,7 +549,15 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
...
         if (errors['__ember_meta__'] && errors['__ember_meta__']['values']) {
-          delete errors['__ember_meta__']['values'][key];
+          let path = key.split('.');
+          if (path.length === 1) {
+            delete errors['__ember_meta__']['values'][key];
+          } else {
+            let branch = path.slice(0, -1).join('.');
+            let [leaf] = path.slice(-1);
+            let obj = get(errors, `__ember_meta__.values.${branch}`);
+            if (obj) delete obj['__ember_meta__']['values'][leaf];
```
nucleartide: Note to self: this may break for triple nested keys, add another test case.

webark:
all the ember meta stuff makes me very nervous... i think we might be able to dig into the relay for this?

nucleartide:
@webark Hmm, would you care to elaborate? It seems like the Ember meta stuff needs to be done for older Ember versions.

addon/index.js
krishandley: #230 Works for nested keys 👍

tests/integration/components/changeset-test.js
```diff
@@ -172,6 +172,75 @@ test('nested object updates when set without a validator', async function(assert
   assert.equal(find('h1').textContent.trim(), 'foo bar', 'should update observable value');
 });
 
+test('nested key error clears after entering valid input', async function(assert) {
```
nucleartide: @krishandley Oh, one more thing – these two tests are nearly the same, I'd like to refactor this into table-driven form like in https://github.com/poteto/ember-changeset/pull/226#discussion_r153973472.

Lemme know if you have time to work on this, if not I'll take a look later today or tomorrow.

krishandley: Yea that’s fine, I can do that.

krishandley: A loop isn't going to work, because variables in the hbs helper don't work.

We can use Ember.HTMLBars.compile, but... we need to add
`app.import('vendor/ember/ember-template-compiler.js');` for 2.13 + and
`app.import('bower_components/ember/ember-template-compiler.js');` for below to the ember-cli-build.js file.

Both do work when running the tests for a single version, but I'm not sure how to target the ember version being used before importing. So I'm at a dead end trying to get that to work, and try/catch didn't work either.

So the simplest solution is just to have multiple inputs, and error divs in the same test/template. If you don't mind that instead?

nucleartide: Hmm, maybe we could set a selector instead of the template string interpolation? For example:

```js
// this is the changeset
this.set('c', c);

// this is test data, where s === 'person.name.parts.first.validation'
this.set('selector', s);
```

```hbs
<input value={{get c selector}} onchange={{action (mut (get c selector))}}>
```

Not completely sure that'll work – would you mind giving it a shot?

krishandley: Good idea, worked like a charm 😄

krishandley: Spoke too soon, 1.13 needs doesn't have the get helper. And adding the get helper addon broke other tests :(

nucleartide: :( which tests are failing?

krishandley: Well, the summary says 8 tests failed, but looking through the only test names which I can see in the failed output are the ones in the loop... But they all say this. Uncaught Error: Stream error: setValue not implemented.

Feel free to take a look at merging the tests yourself, I'll try again tomorrow with fresh eyes if don't get to it.

nucleartide: Just took a look, yeah let's not go down this timesink. 😛

This is fine for now then, great job @krishandley!

webark:
@nucleartide sorry, was out for the weekend. I'll dig into this and see if there's a way we can meet these test cases without needing to dip so far into the private attributes, and if I find a way I'll post a PR.

PR #230, opened 2017-12-03
Added test case for deeply nested error clearing

krishandley:
Tests if a deeply nested error is cleared after entering a valid input (Test passed on 1.1.3)


nucleartide:
Awesome, thanks! 😄

PR #231, opened 2017-12-04
Fix rollback() property notifiers

### LINKED ISSUE #217: changeset.rollback notifies listeners before updating values
Hi,
I have an application with an observer checking the changes made to a changeset field. 
However, when I rollback the changes made and my observer fires, the value is still the old one.
If I delay the code checking the value by placing it in an Ember.run.next then I do have access to the correct rolledback value.

I have created [a twiddle](https://ember-twiddle.com/63299c8032a3dcdcfb231c5a0fdfe4d7?openFiles=templates.application.hbs%2C) to mimic this behavior, just change the firstName, rollback and then you'll see the immediate & delayed value appear.

From this behavior and [what I've seen in the ember-changeset code](https://github.com/poteto/ember-changeset/blob/dcc28a6383577a0674b5eef420819a05c8d5a886/addon/index.js#L246) so far, I suspect that the notifications are sent before the values are actually modified.

krishandley:
Fixes #217

Added keys argument to _notifyVirtualProperties(), so you can pass in
an array of keys instead of always using the changes and error keys.

rollback() has been changed to get the keys first, and then notify the
properties after resetting the changes and errors.

Added _rollbackKeys() private function, so rollback() and
_notifyVirtualProperties() can use the keys in different ways.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -251,7 +251,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
       }
 
       // Get keys before resetting
-      const keys = this._rollbackKeys()
+      let keys = this._rollbackKeys();
 
       set(this, RELAY_CACHE, {});
       set(this, CHANGES, {});
@@ -678,7 +678,7 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
     },
 
     /**
-     * Gets the changes and error keys
+     * Gets the changes and error keys.
      *
      * @private
      * @return {Array}
--- a/addon/index.js
+++ b/addon/index.js
@@ -244,16 +244,20 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap = 
      * @return {Changeset}
      */
     rollback() {
-      this._notifyVirtualProperties();
       let relayCache = get(this, RELAY_CACHE);
 
       for (let key in relayCache) {
         relayCache[key].rollback();
       }
 
+      // Get keys before resetting
+      const keys = this._rollbackKeys()
+
       set(this, RELAY_CACHE, {});
       set(this, CHANGES, {});
       set(this, ERRORS, {});
+      this._notifyVirtualProperties(keys)
+
       return this;
     },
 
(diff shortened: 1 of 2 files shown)


nucleartide:
Nice! Will take a closer look tonight.

addon/index.js
```diff
@@ -244,16 +244,20 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
...
     rollback() {
-      this._notifyVirtualProperties();
       let relayCache = get(this, RELAY_CACHE);
 
       for (let key in relayCache) {
         relayCache[key].rollback();
       }
 
+      // Get keys before resetting
+      const keys = this._rollbackKeys()
```
nucleartide: Use `let` please! just to follow this addon's conventions

nucleartide: Also missing a semicolon. ESLint isn't configured properly it seems.

addon/index.js
```diff
@@ -660,19 +664,32 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
...
-      for (let i = 0; i < rollbackKeys.length; i++) {
-        this.notifyPropertyChange(rollbackKeys[i]);
+    _notifyVirtualProperties(keys = this._rollbackKeys()) {
+      for (let i = 0; i < keys.length; i++) {
+        this.notifyPropertyChange(keys[i]);
       }
     },
 
+    /**
+     * Gets the changes and error keys
```
nucleartide: Nitpick: missing period.

tests/unit/changeset-test.js
```diff
@@ -320,6 +320,18 @@ test('#rollback resets valid state', function(assert) {
   assert.ok(get(dummyChangeset, 'isValid'), 'should be valid');
 });
 
+test('observing #rollback values', function(assert) {
+  let res;
+  const changeset = new Changeset(dummyModel, dummyValidator);
```
nucleartide: Use `let` here as well please!

tests/unit/changeset-test.js
```diff
@@ -320,6 +320,18 @@ test('#rollback resets valid state', function(assert) {
...
+  changeset.addObserver('name', function() { res = this.get('name') });
+  changeset.get('name');
```
nucleartide: This line seems redundant, let's remove.

krishandley:
Updated. A .prettierrc file would be a nice addition to the repo btw ;)

nucleartide:
@krishandley Haha yeah definitely, I think there's something in the latest ember-cli re: prettier. Or maybe it was ESLint or something.

Issue #233, opened 2017-12-05
QA for nested keys release (~1.4.0)


nucleartide:
For all issues, please submit a PR whose initial commit is a **failing test** (as reported by Travis). Subsequent commits should fix the initial failing test.

15. [x] @krishandley `_hasOwnNestedProperty` does not depend on instance state, let's pull this into `utils`.
14. [x] @nucleartide Audit `_setProperty` for possible nested keys bugs.
5. [x] (no tests needed here) Update the JSDoc for `_validate`. `_validate` can also return a Promise.
13. [ ] @nucleartide More detailed JSDoc for `_valueFor`. (What does `plainValue` do?)
16. [ ] @krishandley  `_rollbackKeys` should use the new `leafKey` util. We should also de-dupe the keys here.
10. [ ] @nucleartide `execute` should use `includes` util. See: https://github.com/poteto/ember-changeset/blob/709be66ae6aae9a0a5c1dc21136da2a51c2f2970/addon/index.js#L206
11. [ ] @nucleartide https://github.com/poteto/ember-changeset/issues/229
1. [ ] Ensure `_deleteKey` notifies all properties up the key chain. For example, `_deleteKey(someObj, 'foo.bar.baz')` should notify `foo`, `foo.bar`, and `foo.bar.baz`.
4. [ ] `_setIsValidating` won't work with nested keys. We should use `deepSet` here.
    - Related, `isValidating` should use the new `leafKeys` util.
6. [ ] `cast` may not work with nested keys – write some tests for this.
7. [ ] `pushErrors` will fail if a nested key is passed in. Might need to use `deepSet` here.
8. [ ] `validate` should accept both types of validation maps. Example:
    - `{ 'foo.bar.baz': someValidationFn }`
    - `{ foo: { bar: { baz: someValidationFn } } }`
9. [ ] Ensure `merge` works with nested keys.
12. [ ] Investigate last 2 commits for https://github.com/poteto/ember-changeset/pull/206

krishandley:
Will help if I can!

nucleartide:
@krishandley @webark + anyone who might be reading, I've listed the issues in the original post. Please chip in if you'd like!

webark:
@nucleartide i’ll jump on some of these tomorrow.. thanks!

krishandley:
I'll take an easy one to start with extracting _hasOwnNestedProperty

krishandley:
[...]

PR #234, opened 2017-12-06
Extract _hasOwnNestedProperty to utils

krishandley:
#233

Added some tests, and found an interesting possible bug,
there is one failing test where the last item in the path is true,
which is the value in the object not the key.

### DIFF
--- a/addon/utils/has-own-nested-property.js
+++ b/addon/utils/has-own-nested-property.js
@@ -1,5 +1,12 @@
 import isObject from './is-object';
 
+/**
+ * Checks if a path exists in an object or any of it's children
+ *
+ * @param {Object} object
+ * @param {String} path
+ * @return {Boolean}
+ */
 export default function hasOwnNestedProperty(object, key) {
   let keyParts = key.split('.');
   let nextKey = keyParts.shift();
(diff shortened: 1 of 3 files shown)


nucleartide:
Looks great so far! Thanks for following the table-driven style. 😄

addon/utils/has-own-nested-property.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Checks if a path exists in an object or any of it's children
+ *
+ * @param {Object} object
+ * @param {String} path
+ * @return {Boolean}
+ */
+export default function hasOwnNestedProperty(object, key) {
+  let keyParts = key.split('.');
+  let nextKey = keyParts.shift();
```
nucleartide: Let's shorten to `let [nextKey, ...keyParts] = key.split('.');`

addon/utils/has-own-nested-property.js
```diff
@@ -0,0 +1,35 @@
...
+
+  if (nextKey in object) {
```
nucleartide: Use `object.hasOwnProperty(nextKey)` here to avoid props on the prototype.

nucleartide:
Excellent, just two small issues before merging.

PR #235, opened 2017-12-06
Use leafKeys util instead of _rollbackKeys

krishandley:
#233

_rollbackKeys was getting keys from both the changes and errors before,
but interestingly the tests pass without getting keys from the errors.


krishandley:
Wasn't sure what kind of failing test you wanted for a change like this... So skipped that bit.

nucleartide:
Oh whoops, sorry if I didn't communicate the issue clearly enough.

Basically, I want to switch `_rollbackKeys` to return nested keys as well. Something like this:

```
_rollbackKeys() {
  return [
    leafKeys(get(this, CHANGES)),	// using leafKeys instead of keys	
    leafKeys(get(this, ERRORS))
  ];
}
```

The keys should be in `{ a: { b: { c: { ... } } } }` format because of `deepSet`:

https://github.com/poteto/ember-changeset/blob/709be66ae6aae9a0a5c1dc21136da2a51c2f2970/addon/index.js#L544-L550

krishandley:
The thing that concerns me here is that you can remove `...leafKeys(get(this, ERRORS))` and all the tests still pass.

But I can't figure out how to test for it, it's not very clear what it's doing that `...leafKeys(get(this, CHANGES))` isn't already doing.

nucleartide:
@krishandley This might merit a new test :) try setting a valid value on a changeset, then an invalid value. At this point there should be a one value in `changeset.changes` and one value in `changeset.errors`.

After calling `rollback`, the test should assert that `changeset.changes` is empty and `changeset.errors` is empty.

krishandley:
I tried doing what you suggested, but this created the situation where the keys for changes and errors in _rollbackKeys would have been exactly the same, so it would be impossible to know in a test which keys are being used. And I wasn't trying to test that the values rolled back correctly. Just that they were observed when rolled back.

Please see the commit message with regards to the test I added.

nucleartide:
@krishandley Hmm yeah, I see what you mean.

I wrote up a [small twiddle](https://ember-twiddle.com/56e42ea354363450a845436cc4f95a8d) demonstrating `notifyPropertyChange` behavior. Basically, if your object is `{ foo: { bar: { baz: 4 } } }`, and you call `notifyPropertyChange('foo')`, observers will be fired for `foo`, `foo.bar`, and `foo.bar.baz`. So we don't need to use `leafKeys` - sorry for misleading you.

However, we do need to de-dupe the keys. I believe you can populate `changes` and `errors` with the same key, if you set a valid change first, and then an invalid change. So if the return result of `_rollbackKeys` has duplicates, we will be firing observers more than once.

Lemme know if that makes sense!

krishandley:
Ahh ok, makes sense... but... new problem!

Doing this to create a change and then an error removes the change!

So you can't get duplicate keys this way 😢 

```js
// Valid
changeset.set('name', 'Jack');

// Invalid
changeset.set('name', 'J');

// Changes empty
changeset.get('changes') => []

// _rollbackKeys already unique
_rollbackKeys() => ['name']
```

nucleartide:
@krishandley you’re right, hmm. Let’s add a ‘runInDebug’ assertion to check that the returned array is unique then? It might be the case that changes and errors are always unique now, but if some code changes in the future I’d like for an assertion to catch the possible bug.

krishandley:
I managed to hackily force the changes and error objects to have the same keys to create a failing test. I haven't wrapped that it in a runInDebug check but can do if you want.

addon/index.js
```diff
@@ -692,10 +693,10 @@ export function changeset(obj, validateFn = defaultValidatorFn, validationMap =
      * @return {Array}
      */
     _rollbackKeys() {
-      return [
-        ...keys(get(this, CHANGES)),
-        ...keys(get(this, ERRORS))
-      ];
+      return uniq([
```
nucleartide: You can use Ember's `uniq` here: http://devdocs.io/ember/classes/ember.enumerable/methods#uniq

So something like this:

```js
return emberArray([
  ...keys(get(this, CHANGES)), // don't need leafKeys anymore
  ...keys(get(this, ERRORS))   // don't need leafKeys anymore
]).uniq();
```

nucleartide:
Great work, just a small comment about `uniq`. 🙂

nucleartide:
@krishandley Going to close this PR, sorry. :( I'm making some updates in #254 to address how we represent changes/errors/relays.

Issue #229, opened 2017-12-03
'getting' a Moment instance on a changeset returns a relay instead of that instance


PieterJanVdb:
When having a Moment instance on a changeset, eg.: `set(changeset, 'startDate', moment())`, and if you then later do `get(changeset, 'startDate')` it will - instead of returning the Moment instance on the changeset - return a relay with 'startDate' as key.

I'm unsure if this is a 100% a bug, as the Moment instance is an object so ember-changeset thinks it has to create a relay, but it would be nice if there were a way to deal with this (eg. defining keys for which no relay should be created, as I'm not going to directly get or modify the internals of the Moment instance).

@nucleartide

nucleartide:
Thanks for reporting! This is similar to https://github.com/poteto/ember-changeset/pull/206/commits/a1d22b74bc5ef2fd9d67cd81d628c2382777be20, I'd consider this a bug.

Will include this in the upcoming bugfix release (1.4.2).

nucleartide:
As mentioned on Slack, for now you could try `changeset.get('changeset.someMomentDate')`, since the Relay object stores a reference to `changeset`.

nucleartide:
@PieterJanVdb The latest release [v1.4.2-beta.0](https://github.com/poteto/ember-changeset/releases/tag/v1.4.2-beta.0) fixes this. You can write `changeset.get('someMomentDate')` now and it'll return a Moment instance instead of a Relay.

If you have the time, please try it out (with caution :D). It should still work with `changeset.get('changeset.someMomentDate')` as well.

(Official announcement post is over on [Ember Way](https://emberway.io/ember-changeset-1-4-2-dd9e402d3e10)).

jelhan:
I don't think this is fixed. Here is a reproduction for `v1.5.0-beta.0`: https://ember-twiddle.com/f38bf913c6c44079ba0c84c148cc9f05?openFiles=templates.application.hbs%2C It's the same if you downgrade ember-changeset to `1.4.2-beta.0`.

If accessing the moment object throw changeset (e.g. using `moment-format` helper as in twiddle) this is blowing up due to `.get()` assertion:

```
Assertion Failed: You attempted to access the `_isAMomentObject` property (of <(unknown):ember216>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
[...]

PR #236, opened 2017-12-06
Fix 229 relay objects

krishandley:
#229

### DIFF
--- a/tests/unit/changeset-test.js
+++ b/tests/unit/changeset-test.js
@@ -92,7 +92,7 @@ test('#get returns change that is a blank value', function(assert) {
 });
 
 test('#get returns relay when plain object', function(assert) {
-  let date = {}
+  let date = {};
   let Dummy = EmberObject.extend({});
   let dummyModel = Dummy.create({ date });
   let changeset = new Changeset(dummyModel);
@@ -100,7 +100,7 @@ test('#get returns relay when plain object', function(assert) {
 });
 
 test('#get returns original object when not a plain object', function(assert) {
-  let Moment = () => {}
+  let Moment = () => {};
   let moment = () => new Moment();
   let Dummy = EmberObject.extend({});
   let date = moment();
--- a/tests/unit/changeset-test.js
+++ b/tests/unit/changeset-test.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import Changeset from 'ember-changeset';
 import { module, test } from 'qunit';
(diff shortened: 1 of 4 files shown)


krishandley:
Added a dependency so expecting some rolled eyes 🙄 ;) But a simple fix.

My question is why would you ever want to get the relay object in return?

krishandley:
Hmmm actually I'm not sure this fixes the underlying problem. 🤔

webark:
ya.. devDependencies are cheep, so you could even add moment and test against it directly.

But ya, this is going to be a tough one, cause you never really want to return a relay. You just want to return the object with the updated values. Ideally, in my opinion, the relay, and changeset as a whole for that matter, should always return the same object(copy really) you get, and just be there to track the keys that changed. But that requires a change in usage, or architecture..

But maybe i’m wrong..

tests/unit/changeset-test.js
```diff
@@ -90,6 +90,15 @@ test('#get returns change that is a blank value', function(assert) {
   assert.equal(result, '', 'should proxy to change');
 });
 
+test('#get with nested key returns correct value', function(assert) {
+  let date = {};
+  let foo = { date };
+  let dummyModel = EmberObject.create({ foo });
+  let changeset = new Changeset(dummyModel);
+  assert.equal(changeset.get('foo'), foo);
+  assert.equal(changeset.get('foo.date'), date);
```
webark: I think this should be a deep equals here

webark:
and I think this is the relevant test case that got brought up during the development of nested keys.

https://github.com/poteto/ember-changeset/blob/709be66ae6aae9a0a5c1dc21136da2a51c2f2970/tests/unit/changeset-test.js#L933-L952

tests/unit/changeset-test.js
krishandley: I wanted to check for the exact object, so equal was what I was going for.

krishandley:
Closing. Original fix did work but need to dig deeper to understand the nesting better.

tests/unit/changeset-test.js
webark: ya, but `{} === {}` is false.

krishandley: That's not what was being tested. foo is already known. `foo === foo` would be true. It's not testing a new empty object against another empty object.

webark: 🤦‍♂️ yes. Sorry. ... 🐑 ...

PR #237, opened 2017-12-08
Fix: delete nested key on `.set()`

nucleartide:


ynnoj:
@nucleartide Is there a patch release incoming for this fix?

nucleartide:
@ynnoj Will look into releasing tonight, I just figured everyone was on vacation for the past week..

ynnoj:
@nucleartide Appreciate that! back to the grind now. Happy New Year 🎆

PR #240, opened 2017-12-09
Fix: #execute should not remove original nested objects

nucleartide:


nucleartide:
Hmm, I can't get #3 to pass. Table-driven tests #1 and #2 return Relay instances so they work, but not #3.

nucleartide:
Duplicate of #258.

PR #241, opened 2017-12-11
Add alternative use cases to README

sukima:
I found these two patterns very useful when working with changesets I felt other readers would appreciate them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,6 +675,34 @@ if (isChangeset(model)) {
 - [`ember-changeset-hofs`](https://github.com/nucleartide/ember-changeset-hofs) - Higher-order validation functions
 - [`ember-bootstrap-changeset-validations`](https://github.com/kaliber5/ember-bootstrap-changeset-validations) - Adds support for changeset validations to `ember-bootstrap`
 
+## Alternative Use Cases
+
+If you wish to use a changeset from a template without a component you can use the `with` helper:
+
+```hbs
+{{#with (changeset model (action "validate")) as |changeset|}}
+  Name: <input value={{changeset.name}}>
+  <button {{action "saveThings" chageset}}>Save</button>
+{{/with}}
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -675,6 +675,34 @@ if (isChangeset(model)) {
...
+```js
+changeset: computed('model', {
+  get() {
+    return new Changeset(this.get('model'));
+  }
+}),
+
+save() {
+  let changeset = this.get('changeset');
+  doSomethingFancyWith(changeset);
```
nucleartide: Could you elaborate a little more on this `doSomethingFancyWith` function? I'm curious why we would want to create a new changeset entirely.

README.md
```diff
@@ -675,6 +675,34 @@ if (isChangeset(model)) {
 - [`ember-changeset-hofs`](https://github.com/nucleartide/ember-changeset-hofs) - Higher-order validation functions
 - [`ember-bootstrap-changeset-validations`](https://github.com/kaliber5/ember-bootstrap-changeset-validations) - Adds support for changeset validations to `ember-bootstrap`
 
+## Alternative Use Cases
+
+If you wish to use a changeset from a template without a component you can use the `with` helper:
```
nucleartide: This usage is already (somewhat) suggested at https://github.com/poteto/ember-changeset#disabling-automatic-validation - I'd rather not bloat the readme. :(

sukima: Is there a way to make that more explicit? I've read this README several times and still didn't glean that use case. Perhaps I just skipped over that because I wasn't thinking about validations which that section suggests is tied to `{{with}}`. I think the implied usage is confusing it would help if it wasn't tied to validations. Thoughts?

nucleartide:
@sukima We could start using the wiki for documenting use cases. I'll let you know - currently the README is in flux, because I'm thinking about nested keys docs.

README.md
nucleartide: Updated here: https://github.com/poteto/ember-changeset/tree/develop/nested-keys#usage

PR #243, opened 2017-12-12
Update imports to use @ember packages

bbtb1982:
# What
Update all references to use @ember packages with named exports and alias as noted in [ember-rfc176-data](https://github.com/ember-cli/ember-rfc176-data)

# Why
Lift the documentation and code base to more inline with v2.16+ emberJS docs

# Test
npm test

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,4 +1,4 @@
-import Ember from 'ember';
+//import Ember from 'ember';
 import Relay from 'ember-changeset/-private/relay';
 import objectToArray from 'ember-changeset/utils/computed/object-to-array';
 import isEmptyObject from 'ember-changeset/utils/computed/is-empty-object';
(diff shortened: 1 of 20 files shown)


addon/index.js
```diff
@@ -1,4 +1,4 @@
-import Ember from 'ember';
+//import Ember from 'ember';
```
nucleartide: Remove comment please!

addon/utils/computed/object-equal.js
```diff
@@ -1,6 +1,5 @@
-import Ember from 'ember';
+import { computed, get} from '@ember/object';
```
nucleartide: Nitpick: add a space between `get` and `}`.

bbtb1982:
@nucleartide ready for review

PR #247, opened 2017-12-13
Fix leaf keys with promises

krishandley:
When leafKeys encountered a promise, there would be lots of extra keys from the promise object like x._id and about 5 others. This fixes that. Needed for #233

### DIFF
--- a/addon/utils/leaf-keys.js
+++ b/addon/utils/leaf-keys.js
@@ -1,4 +1,5 @@
 import isObject from 'ember-changeset/utils/is-object';
+import isPromise from 'ember-changeset/utils/is-promise';
 import { isArray } from '@ember/array';
 
 /**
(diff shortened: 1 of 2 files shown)


nucleartide:
I've been thinking - we should revise the "schemas" for our internal objects: 

https://github.com/poteto/ember-changeset/blob/a7652bfa5585690f01fceedd0605085dd09aa690/addon/index.js#L83-L92

`CHANGES`, `ERRORS`, `RELAY_CACHE`, and `RUNNING_VALIDATIONS` are currently nested objects - which works fine - but we need a way to "whitelist" the change/error/relay/validation instead of blacklisting things that aren't. For example:

```js
// this is a change object
{ __struct__: CHANGES, value: 'some change' }

// this is also a change object
{ __struct__: CHANGES, value: Promise.resolve(42) }

// same here
{ __struct__: CHANGES, value: moment() }

// this is the CHANGES map
{
  foo: {
    bar: {
      hello: { __struct__: CHANGES, value: Promise.resolve(42) },
      world: { __struct__: CHANGES, value: moment() }
    }
  }
}
```

(I think the `__struct__` is appropriate since this _is_ an Elixir inspired lib. 😛 )

This way, running `leafKeys` on the `CHANGES` map (or any map) will be a lot more robust.

nucleartide:
@krishandley Thanks a bunch for the PR by the way, sorry if I might have misled you. Lemme update #233.

krishandley:
No worries, feel free to close this and make an issue for restructuring.

nucleartide:
Thanks a bunch @krishandley, will address this as part of #254.

PR #249, opened 2017-12-15
Add nested keys docs

### LINKED ISSUE #228: Docs for nested key support
TODO: may need to explain `Relay` cache.

nucleartide:
If anyone has the time, please take a look at these updated docs for nested keys! Nitpicks, wording suggestions, or anything at all would be appreciated.

I'll also take a second look at this after getting some sleep.

@sukima Re: https://github.com/poteto/ember-changeset/pull/241, I'm thinking we could change the first code example to something like this?

```hbs
{{#with (changeset model (action "validate")) as |c|}}
  {{dummy-form changeset=c submit=(action "submit") rollback=(action "rollback")}}
{{/with}}
```

Preview: https://github.com/poteto/ember-changeset/tree/docs

Closes #228.


nucleartide:
@poteto @krishandley @webark @sukima

sukima:
I'm ok with this. What do you need me to do? Should I close #241 ?

nucleartide:
@sukima Oh, feel free to open a new PR for the `{{with}}` docs change. I don't want to take your contributor credit haha.

For the `doSomethingFancyWith` docs, could you open another issue? I think that's something we should fix in changeset.

webark:
`errors` comes flat like that? And I know you were talking about changing it, but if you define your own validation, you have to nest them, not just have a `nested.key.name`

nucleartide:
@webark Yeah `errors` has always been a flat array, I figured nesting wouldn't make sense.

I also think we should accept both styles for validations; `nested.key.name` and `nested: { key: { 
name } }` should both work.

webark:
ok cool. Looks good to me then!

nucleartide:
Going to merge this. Thanks for the look-over @webark! 😄

PR #258, opened 2017-12-19
Refactor #execute method

nucleartide:


nucleartide:
Will fix failing tests as part of `refactor/value-for`.

webark:
ya. This method got ugly when I needed to use `setProperties` to trigger "changed attributes" on ember-data models.

PR #262, opened 2017-12-19
Add Flow comment types

nucleartide:
I introduced Flow comment types to my workplace, and I figure they'd be useful here. Some reasons:

1. Flow comment types require no build tooling. You just need to install `flow`, annotate your file with `// @flow`, and write all your types within `/*: */` or `/*:: */` comments.
2. We could try to mirror the types in `validated-proxy`.
3. I don't think `ember-cli-typescript` works in addons yet? Might be wrong.
4. With the new `Change` and `Err` representations, it's hard to mentally keep track of types; Flow does this for us automatically. It'd help track down subtle type errors.

@poteto @webark @krishandley et al. – whaddya think?


nucleartide:
Going to merge and work on this as part of #254.

Having types clarified a lot of implementation questions. My thinking is that our internal `CHANGES`, `ERRORS`, `RELAY_CACHE` maps, etc. should always be one level deep: nested keys are stored as is, like `{ 'foo.bar.baz': someChange }`. Public properties like `error` and `change` then compute their result from our internal representation.

This means we don't even need `ember-deep-set`, unless I'm missing something.

Issue #256, opened 2017-12-18
Support nested keys


nucleartide:
`ember-changeset` 1.4.0+ adds support for nested keys. However, the latest minor release (1.4.1) also breaks existing APIs, and introduces bugs because not all `Changeset` methods have been updated to support nested keys.

Thus, **you should not use `ember-changeset` 1.4.0+, nor should you point your `package.json` at the latest commit on `master`**. Instead, please pin your `ember-changeset` version to **1.3.0**:

```js
{
  "dependencies": {
    "ember-changeset": "1.3.0"
  }
}
```

If you're using `ember-changeset-validations` (e-c-v), you should also pin e-c-v's version to **1.2.8**:

```js
{
  "dependencies": {
    "ember-changeset": "1.3.0",
    "ember-changeset-validations": "1.2.8"
  }
}
```

The [`develop/nested-keys`](https://github.com/poteto/ember-changeset/tree/develop/nested-keys) branch contains ongoing work on nested keys support. We'll make a formal announcement when a bugfix release is ready.

nrcjbarry:
Thank you! I just ran into these issues last week. All our environments (Dev/Test/Stage) were throwing exceptions in one form's save, but I couldn't reproduce locally. Ended up figuring this out after doing multiple `rm -rf node_modules` and `npm installs`.

sdhull:
Ah! I didn't see this. Thanks for pointing me here.

First: it should go without saying (however I'll say it anyway!) I very much appreciate the addon and all the contributors' work it took to make it happen.

Second: I didn't see this before coming up with a workaround, which is to simply do:
```javascript
object.setProperties(get(changeset, 'change'))
object.save()
```
Which seems to be working OK so far.

Last: I skipped over this issue because the title (`Support nested keys`) didn't strike me as applying to my problem.

Suggestions: I would recommend updating this title ("Support nested keys: 1.4.0-1 broken") or maybe even updating the Readme to reflect the problems with those versions. Or ... is it possible to unpublish versions from npm? That might be a good idea.

PR #271, opened 2017-12-25
Update README.md

scathers:
Remove outdated version number from ember install command


nucleartide:
Hi @scathers, thanks for the PR. The version is pinned at 1.3.0 for a reason, please see https://github.com/poteto/ember-changeset/issues/256.

PR #279, opened 2018-01-27
Update ember-deep-set dependency

jbescoyez:
...in order to remove the deprecation notice

See: https://github.com/poteto/ember-deep-set/pull/10
And: https://github.com/nucleartide/ember-changeset-hofs/issues/1

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "homepage": "https://github.com/poteto/ember-changeset",
   "dependencies": {
     "ember-cli-babel": "^6.8.2",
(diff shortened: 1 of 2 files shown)


nucleartide:
@jbescoyez Just published `ember-deep-set@0.1.3`, would you mind updating the `package.json` here to `^0.1.3` instead?

jbescoyez:
@nucleartide Here it is.

nucleartide:
@jbescoyez Thank you!

PR #280, opened 2018-01-29
belongsTo properties not being changed - always reverts to original object

endevea:
If you change a belongsTo property on a changeset to a new object then it does not stick. All "get" calls revert back to the original object. 

This PR redirects any "get" calls to use the object stored in the _changes object (if it exists), instead of what is stored in the relayCache.

Unfortunately, this change breaks some unit tests: "table driven test - X".  Looking at the tests, it is not clear if the tests are meant to work if the object is changed i.e.

The second test sets the "org" property to be "no usa for you", but then assumes that the "org" property still points to the old object when it sets "org.usa.ca". This doesn't seem to match what people expect if they change the "org" property.

Any guidance would help as I'd like to submit this fix cleanly, but am unsure what to do next.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -698,13 +698,14 @@ export function changeset(
       }
 
       let original /*: mixed */ = get(content, key);
-      if (isObject(original) && !plainValue) {
+      let hasChanged /*: boolean */ = changes.hasOwnProperty(key);
+      if (isObject(original) && !plainValue && !hasChanged) {
         let c /*: ChangesetDef */ = this;
         let o /*: Object       */ = (original /*: any */);
         return c._relayFor(key, o);
       }
 
-      if (changes.hasOwnProperty(key)) {
+      if (hasChanged) {
         let c /*: Change */ = changes[key];
         return c.value;
       }


nucleartide:
Thanks for the PR – I will take a look at the failing tests later tonight. If this is indeed a bug, then we should add another test for the expected behavior as well.

fran-worley:
@nucleartide I'm experiencing exactly this problem, if I set the relation on the model it sticks and we're happy but if I set it on the changeset it simply reverts back to what it was.

If you need help producing a replicable example let me know but I'm +1 this as a bug.

nucleartide:
Yes this is a bug then, thanks for the catch! @fran-worley or @endevea, if you get the time, would you mind adding an additional failing test for the expected `belongsTo` setting behavior, then try to make it pass? Please call dibs so it's clear who's working on it. :)

@endevea, the test that you pointed at tests the `model`, not the `changeset`. So that test is properly failing – we'll need to fix that as well. For clarity, we could improve that group of tests by adding test descriptions if you'd like.

fran-worley:
@endevea @nucleartide this fails before the change in this PR and passes after:

```javascript
// file: ember-changeset/tests/acceptance/main-test.js
test('it works for belongsTo relation', function(assert) {
  let store = this.application.__container__.lookup('service:store');
  let user = this.dummyUser;

  run(() => {
    let anotherProfile = store.createRecord('profile', { firstName: 'Joe', lastName: 'Blog' });
    let changeset = new Changeset(user);
    
    assert.notEqual(anotherProfile.get('firstName'), user.get('profile.firstName'));
    changeset.set('profile', anotherProfile);
    assert.equal(changeset.get('profile.firstName'), anotherProfile.get('firstName'));

    changeset.execute();

    assert.equal(user.get('profile.firstName'), anotherProfile.get('firstName'));
  })
});
```

fran-worley:
For good measure we could also add in some tests against adding and removing hasMany associations as well:

This two pass before and after the proposed change but would improve test coverage around relations.

```javascript
test('it works for removing hasMany relation', function(assert) {
  let user = this.dummyUser;

  run(() => {
    let changeset = new Changeset(user);

    assert.equal(2, changeset.get('dogs.length'));
    changeset.get('dogs').popObject();
    assert.equal(1, changeset.get('dogs.length'));

    changeset.execute();

    assert.equal(1, user.get('dogs.length'));
  })
});

test('it works for belongsTo relation', function(assert) {
  let store = this.application.__container__.lookup('service:store');
  let user = this.dummyUser;

  run(() => {
    let anotherProfile = store.createRecord('profile', { firstName: 'Joe', lastName: 'Blog' });
    let changeset = new Changeset(user);
    
    assert.notEqual(anotherProfile.get('firstName'), user.get('profile.firstName'));
    changeset.set('profile', anotherProfile);
    assert.equal(changeset.get('profile.firstName'), anotherProfile.get('firstName'));

    changeset.execute();

    assert.equal(user.get('profile.firstName'), anotherProfile.get('firstName'));
  })
});
```

nucleartide:
@fran-worley Awesome, and yes please open a PR!

endevea:
Sorry, been on the road. Thanks for picking this up @fran-worley

fran-worley:
@nucleartide unfortunately due to a deadline I've had to move away from this addon for the project I'm currently working. Even with the fix proposed here a number of weird bugs appeared around assigning relations (both belongs-to and has-many) that simply disappeared when I removed changesets. 

Sadly right not I've just not got the time to delve into the code and workout exactly what the problem is so a PR from me probably won't be much good to you! If I get anymore time I'll revisit and see if I can help.

nucleartide:
@fran-worley Ah bummer, thanks for raising the issue though! We definitely need more test coverage around Ember Data relations. It seems like ED models don’t interact well with the relay cache implementation for nested keys.

If anyone is interested, please feel free to add some test coverage and open a PR. I’ll also look into this if time permits.

snewcomer:
I'll jump on this one this weekend. 👍

PR #288, opened 2018-02-22
ember-cli-update to 3.0

Dhaulagiri:
Updates everything via ember-cli-update to 3.0.

Note that I removed ember-1.13 from the testing matrix because it refused to boot the test app for acceptance tests under the new test setup.  If this is problematic I _should_ be able to undo that particular change, however at this point I'm not sure how much support needs to exist for apps on 1.13 still.


Dhaulagiri:
@snewcomer turns out I can hit the button here. I think the only reason I hesitated was the fact that I removed 1.13 from the testing matrix. To me it seems minor but I wanted a sanity check from someone else that this wasn't going to feel like a potential breaking change?? 🤷‍♂️

snewcomer:
Oh yeah.  Just saw that you were a maintainer.  hm yeah understandable.  Perhaps in the interest of future development, it might be a good idea.  `1.3.0` is stable for anybody that would like to stick to an older version of Ember; however, seems a bit of work to get a few items back in working order (relationships).  I would be in favor of merging if you feel comfortable.

Dhaulagiri:
Yeah at this point I think 1.13 fans could stick with 1.3.0 if they had to

Issue #178, opened 2017-04-26
Question: Is it possible to extend Changeset class ?


janmisek:
I am looking into possibility to add custom computed properties on changeset.  Is it possible to extend Changeset class?

Thanks in advance.

SuperCoolFrog:
The addon exports the class which is an ember object that you can extend on.  
I wrote this quick twiddle but it should point you in the right direction.
[Twiddle](https://ember-twiddle.com/e8db2f5a2165942e96e9378be3ebba57)

Extend API Guide
https://emberjs.com/api/classes/Ember.Object.html#method_extend

janmisek:
Thanks alot for explanation I found myself yesterday. 

I would also suggest to include this info in readme, because export of `changeset` class is nowhere documented, default export is "ember-non-extensible" constructror function. 

Btw: why the original ember class is wrapped to classic js constructor function exported by default - is it to make original class semi-private or the reason is e.g typescript compat.

bgentry:
The example in that Twiddle doesn't actually work. If you use that method, then the changeset's `obj`, `validationMap`, etc are all fixed to the default values and there is no way to change them. It does not appear that the Changeset class is actually extensible.

Is that intentional? Does it need to be that way?

snewcomer:
We should ensure this works with 3.0.0....

snewcomer:
This is now documented in the README!

PR #292, opened 2018-03-24
add an afterRollback event

bgentry:
## Add an `afterRollback` event hook

This adds a new event hook `afterRollback` that gets called at the end of a changeset `rollback()`. I wouldn't have a PR for this if I could cleanly extend the `Changeset` class myself, but, I can't: https://github.com/poteto/ember-changeset/issues/178#issuecomment-375852516

This is required for me to have a reasonably clean solution to https://github.com/offirgolan/ember-changeset-cp-validations/issues/25 . I'll post more info there to help explain why I need this.

Cheers :v:

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ Be sure to call `validate()` on the `changeset` before saving or committing chan
 * Events
   + [`beforeValidation`](#beforevalidation)
   + [`afterValidation`](#aftervalidation)
+  + [`afterRollback`](#afterrollback)
 
 #### `error`
 
@@ -739,6 +740,22 @@ changeset.validate().then(() => {
   // console output: lastName has completed validating
   // console output: address.city has completed validating
 });
+```
+
+**[⬆️ back to top](#api)**
+
+#### `afterRollback`
+
+This event is triggered after a rollback of the changeset.
+This can be used for [some advanced use cases](https://github.com/offirgolan/ember-changeset-cp-validations/issues/25#issuecomment-375855834)
+where it is necessary to separately track all changes that are made to the changeset.
+
+```js
+changeset.on('afterRollback', () => {
+  console.log("changeset has rolled back");
+});
+changeset.rollback();
+// console output: changeset has rolled back
 ```
 
 **[⬆️ back to top](#api)**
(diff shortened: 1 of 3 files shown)


bgentry:
Any reason not to merge this? Let me know if you want any changes.

README.md
```diff
@@ -743,6 +744,20 @@ changeset.validate().then(() => {
 
 **[⬆️ back to top](#api)**
 
+#### `afterRollback`
+
+This event is triggered after a rollback of the changeset.
```
snewcomer: @bgentry Would you mind adding a few sentences detailing the "use case" for users?  To me it isn't immediately obvious, but I trust that there is a valid use case.

addon/index.js
```diff
@@ -396,6 +397,7 @@ export function changeset(
       set(this, ERRORS, {});
       (this /*: ChangesetDef */)._notifyVirtualProperties(keys)
 
+      this.trigger(AFTER_ROLLBACK_EVENT);
```
snewcomer: I think to follow some other places in the code, one can do:

```
let c          /*: ChangesetDef     */ = this;
c.trigger(BEFORE_VALIDATION_EVENT, key);
```

bgentry:
@snewcomer let me know if these changes are sufficient. The best I could do to explain it is to link to an existing post where I described why I needed this functionality.

I think it is somewhat related to the larger issue discussed in https://github.com/poteto/ember-changeset/issues/297, which is that the changeset does not actually represent the most recently-set values for all fields (mainly those that are invalid or haven't yet completed validations).

snewcomer:
@bgentry I think if you merge master in, we should get a passing test suite!  The comment you added looks good enough from my end 👍

bgentry:
@snewcomer rebased as requested :v:

snewcomer:
@bgentry there might a `flow` error or something.  Not showing the exact issue in the travis fail though

bgentry:
@snewcomer yeah, it was a flow error because the new callback does not have a 2nd string arg (i.e. the key of the attr being changed). I changed the flow signature to `string | void`, hopefully that's the correct choice.

snewcomer:
@Dhaulagiri Mind taking a look at this PR?  Increases the API surface, but does provide a valuable event for some use cases.

snewcomer:
@Dhaulagiri Lmk if you have time to look at these PR's today.  Going to tackle a much lgr feature after this is in ☮️

Issue #297, opened 2018-05-30
long-running promise validations cause UI delays, unpredictable results


bgentry:
## Version

v1.5.0-beta.0

## Test Case

I believe I've uncovered a flaw in the design of this addon around how it handles promise-based validations. Take the following `<input >` wired up to a changeset:

```hbs
<input value={{changeset.email}} oninput={{action (mut changeset.email) value="target.value"}}>
```

Simple scenario. Now imagine you're validating that email against a remote server, i.e. the promise will take awhile and the duration will of course be variable.

When you set a property on the changeset, it calls [its `_validateAndSet` method](https://github.com/poteto/ember-changeset/blob/353d0e5822efca3104a2b147e47608bc0176e440/addon/index.js#L574-L604). For validations that return promises, the following happens:

1. User calls `changeset.set('email', "john@example.com")`
2. Validation returns a promise, which can be as simple as a timeout / sleep to illustrate this issue
3. The promise is resolved by `_validateAndSet`, anywhere from a few ms to several seconds
4. The validation result is then passed to `_setProperty`
5. `_setProperty` actually sets the new value on the changeset, along with any validation error

The entire time this is happening, the input continues to render with the old value because `changeset.email` has not actually changed yet (not until the end of step 5). So I could type `j` and then continue typing the rest of the email, but the input will appear to be frozen. Once step 5 finishes (which could be hundreds or thousands of ms), then the field actually updates.

This gets even worse if you have multiple validation requests in-flight, which will probably happen even with debouncing. You're not actually guaranteed that any of these promises will resolve in the same order they were started, so you can potentially get the `joh` promise resolving before the `jo` promise, in which case `changeset.email` will ultimately be set to `jo`.

I hope I'm seriously misunderstanding something here because it doesn't seem like it will be easy to fix 😕

bgentry:
I think I have a promising lead on solving this! My solution involves a couple of key changes:

* rather than `_validateAndSet`, do `_setAndValidate`. Immediately set the new value on `_changes` before starting the validation.
* Validate using ember-concurrency, ensuring that only one validation of each property is running at a time. I created a `PropertyValidator` object type for this:

```js
const PropertyValidator = EmberObject.extend({
  key: null,
  validator() {},

  isRunning: readOnly("validate.isRunning"),

  validate: task(function*(changeset, newValue, oldValue, changes, content) {
    let result = yield this.validator({
      key: this.key,
      newValue,
      oldValue,
      changes,
      content,
    });
[...]

PR #303, opened 2018-07-11
Ensure error changes are added to changes block

snewcomer:
I haven't investigated the implications of this change; however, this is what it would look like.  Note - there are still tests that `execute` and do not apply those changes if invalid.  

I'm not sure if there is historical reasons for this or if simply was a design decision.

This also might require more information in the README.

Ref #297 
https://github.com/poteto/ember-changeset/issues/297#issuecomment-400101621

### DIFF
--- a/README.md
+++ b/README.md
@@ -902,7 +902,9 @@ export default Component.extend({
     }
   }
 });
+```
 
+```hbs
 <input
   type={{type}}
   value={{get model valuePath}}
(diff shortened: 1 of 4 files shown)


snewcomer:
@bgentry Mind trying out this PR and lmk if anything breaks for you (of course if you have time :) )?

snewcomer:
@XaserAcheron Somewhat modifying what you did (the tests), but would love to hear your opinion on this PR.  It essentially still adds `changes` to the changeset even if it is not valid.

XaserAcheron:
This breaks `rollbackInvalid`, but it ought to be a simple fix. You'll want to revert the definition of the test and add a couple of lines [here-ish](https://github.com/poteto/ember-changeset/blob/2bd3e0f5127cee9461485447ffbe63c23eeee657/addon/index.js#L420) to also remove the invalid items from the `CHANGES` hash. Should do the trick.

snewcomer:
@XaserAcheron ooo great call!  Hadn't thought of that.  Since the errors are `potential`, we need to see if they actually exist and then delete from the changes hash if present in the errors hash.  👍

XaserAcheron:
Just saw the commit -- looks good now! Or at least that little tiny feature I contributed looks good now, for what it's worth. :P

PR #306, opened 2018-07-13
Update to 3.2

snewcomer:


snewcomer:
@Dhaulagiri Mind doing a quick review when you get a chance?  Need this for other PR's in the pipeline. 🙏

.travis.yml
```diff
@@ -39,20 +36,14 @@ matrix:
...
-  - yarn install --no-lockfile
-
-before_script:
-  - npm rebuild node-sass
+  - yarn install --no-lockfile --non-interactive
 
 script:
-  - yarn flow
-  - npm run lint:js
+  - yarn lint:js
```
Dhaulagiri: Do we still need to run flow here?

package.json
```diff
@@ -27,36 +27,36 @@
...
     "eslint-plugin-ember": "^5.0.0",
-    "eslint-plugin-node": "^5.2.1",
+    "eslint-plugin-node": "^6.0.1",
     "flow-bin": "^0.61.0",
     "flow-typed": "^2.2.3",
     "loader.js": "^4.2.3"
   },
   "engines": {
-    "node": "^4.5 || 6.* || >= 7.*"
+    "node": "6.* || 8.* || >= 10.*"
```
Dhaulagiri: In theory dropping node 4 support is a breaking change so we should probably leave that

snewcomer: Since we are on a beta branch, do you think its ok to leave out?  I think node 4.* only supports up to 3.1 as well.  Lmk what you think.

https://github.com/ember-cli/ember-cli/pull/7801/files

Dhaulagiri: 🤷‍♂️  I don't have a strong feeling about it since my experience is that I haven't used node 4 for a very long time, but I know I've seen other addons be cautious about this.

snewcomer: @Dhaulagiri Yeah I think you are right.  I put the `^4.5` back 👍

PR #307, opened 2018-08-08
rollback invalid key addition

### LINKED ISSUE #282: Feature Request: A way to remove errors without rolling back
In our form component and app we have a concept of `non_field_errors` which get returned from the api (default django-rest-framework behavior). These errors are "form-level" and not tied to a specific field.

To keep things simple we use our changeset form to display these if they occur by doing something like:
```
        try {
            yield bankcard.save();
        } catch (err) {
            err.errors.forEach(({ attribute, message }) => {
                changeset.pushErrors(attribute, message[0]);
            });
            return undefined;
        }
```
which if `err` === `non_field_errors`, our form template will render at the top of the form like:
```
    {{cool-form/error changeset "non_field_errors"}}
```
The problem is if a user submits and receives a non_field_error, we have to then handle clearing just that one specific error before they can submit again. While this is easy to do, to delete just that specific error with no related input attached we have to do `delete changeset._errors.non_field_errors`

Proposal: Can we add a public interface to remove a single error without relying on private properties?

e.g. `changeset.removeError('non_field_errors')` or `changeset.clearError('non_field_errors')`

OR just a public way to clear all errors e.g. `changeset.clearErrors()` without rolling the form back

snewcomer:
@erichonkanen @XaserAcheron Lmk what you think!

close #282

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -434,7 +434,7 @@ export function changeset(
      * @param {String} key key to delete off of changes and errors
      * @return {Changeset}
      */
-    rollbackProperty(key) {
+    rollbackProperty(key) /*: ChangesetDef */ {
       this._deleteKey(CHANGES, key);
       this._deleteKey(ERRORS, key);
 
--- a/addon/index.js
+++ b/addon/index.js
@@ -150,7 +150,7 @@ export type ChangesetDef = {|
   _notifyVirtualProperties: (?Array<string>) => void,
   _rollbackKeys: () => Array<string>,
   rollback: () => ChangesetDef,
-  rollbackInvalid: () => ChangesetDef,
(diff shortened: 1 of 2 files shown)


XaserAcheron:
Cool! That looks useful for sure.

I feel like my 2c is a bit minor since I'm just a 3rd-party contributor, but there it is, for what it's worth. Thanks for the tag!

XaserAcheron:
Looking at it though, the Flowtype def on [line 153](https://github.com/poteto/ember-changeset/blob/ebafb2113f2ec76aa296ff61d8ae9fec6769367e/addon/index.js#L153) probably needs updating.

The rollbackInvalid and rollbackProperty functions themselves are missing Flow types, though that's mostly 'cause the project didn't use Flow back when I made the PR. Ah well.

Issue #315, opened 2018-09-08
belongsTo not properly emptying when set to `null`


nickschot:
## Version

1.5.0

## Test Case

#314 

## Steps to reproduce

I first spotted this in a changeset powered ember-bootstrap form w/ a power-select for (re)setting a belongsTo relationship. Works fine on 1.3.0.

## Expected Behavior

The relationship is emptied properly.

## Actual Behavior

It is not emptied :(

nickschot:
Found the piece of code that causes this:

https://github.com/poteto/ember-changeset/blob/7bda87b2ae8f07754f7d07a122fe2efc4800918f/addon/index.js#L767-L771

nickschot:
@snewcomer is there a need to relay anyway if the new value is not an object (mainly `null`)?

nickschot:
@snewcomer this seems to have broken again for me on 1.6.0 even though the PR #318 is in that release.

snewcomer:
What release did it last work in?

nickschot:
1.3, it seems #318 also didn't fix this exact issue as the section of code that's triggered is above that (see a couple comments back). I think the problem is that it only checks if the original is an object and then returns the relay, while the true value is now `null` as it's an empty relationship.

snewcomer:
I’m guessing this is fixed now as of 2.0z. I’ll close for now but feel free to reopen!

PR #314, opened 2018-09-08
add failing tests for nulling belongsTo relationship

nickschot:
See #315 .

## Changes proposed in this pull request

Added some failing tests for deleting a belongsTo relationship on a changeset.

PR #316, opened 2018-09-11
Move around tests and use new qunit apis

snewcomer:


snewcomer:
@nickshot is this the same error you were seeing?  My initial check was just to move it out of a 

> TypeError: (0 , _emberMetal.get)(...).getURL is not a function

If so, we might have to chalk this up as one other case where `Relay` has gone wrong for us.
 
Ref #315

nickschot:
That is the error that's thrown in the failing test (#314) indeed. I do not see this error in my bootstrap-form (w/ changeset & ember-model-select). Could be that it's caught somewhere.

PR #327, opened 2018-11-15
Convert addon to typescript

dorilla:
Closes #323

## Changes proposed in this pull request
This PR removes the Flow implementation and converts to Typescript.

## TODO
- [x] convert `index.js`
- [ ] convert tests
- [ ] add tslint
- [x] fix https://github.com/poteto/ember-changeset/pull/327#discussion_r233720588
- [x] fix https://github.com/poteto/ember-changeset/pull/327#discussion_r233720989
- [x] fix https://github.com/poteto/ember-changeset/pull/327#discussion_r233721180

Issue #317, opened 2018-09-11
RFC: v2.0


snewcomer:
This issue is to track looking into and potentially removing the Relay [implementation](https://github.com/poteto/ember-changeset/blob/master/addon/-private/relay.js).  I was not around when this was implemented but to the best of my knowledge, when trying to set a nested property/get a nested property, the Ember proxy is recursively called with each successive key.  

GIven that we can't have native property accessors on the changeset (nor `Ember.get(`), this has cause considerable pain, especially has Ember core has moved to allowing native ES5 descriptor variant of `get`.  See issues below.  I have not found a way around this.

#315 
#281 

**Action steps**
I would advocate for a simpler model either removing the Relay implementation and/or involving native (polyfilled) Proxy in combination with https://github.com/poteto/ember-deep-set.  It is hard to know what it looks like in the end until we play with it some more.

We have a nice suite of tests to fall back on for this refactor.  However, I'm not sure all the edge cases `unknownProperty` helps us with, so that is the one outstanding item we need to be aware of.

Some Relay PR's.
https://github.com/poteto/ember-changeset/issues/242
https://github.com/poteto/ember-changeset/pull/194/files

Ref #301 #245 #242 

I will be really busy for the next month or so, but I'll see what I can dig into.  If not much traction is gained/nobody else takes up this task, be sure it will be implemented by the end of October.

snewcomer:
Just an update so far!  `2.0-beta branch created

Pt. 1 merged into `2.0-beta` branch.
https://github.com/poteto/ember-changeset/issues/322

Pt. 2 a work in progress by @dorilla!

PR #328, opened 2018-11-22
Use deepSet when updating RUNNING_VALIDATIONS for nested key

SergeAstapov:
When ember-changeset is used for complex objects with nested keys in combination with [ember-cp-validations](https://github.com/offirgolan/ember-cp-validations) (via [ember-changeset-cp-validations](https://github.com/offirgolan/ember-changeset-cp-validations)) changes to nested keys produce an error `Property set failed: object in path "person" could not be found or was destroyed.`

Reproduction can be found at https://9j9o487k3o.sse.codesandbox.io/ (open console and click on checkbox).

Use `deepSet` instead of `set` to support nested keys.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -744,7 +744,7 @@ export function changeset(
         return;
       }
 
-      set(running, key, value ? count+1 : count-1);
+      deepSet(running, key, value ? count+1 : count-1);
     },
 
     /**
--- a/tests/integration/components/changeset-test.js
+++ b/tests/integration/components/changeset-test.js
@@ -210,6 +210,28 @@ module('Integration | Helper | changeset', function(hooks) {
     }
   });
 
+  test('nested object updates when set with async validator', async function(assert) {
+    let data = { person: { firstName: 'Jim' } };
+    let validator = () => resolve(true);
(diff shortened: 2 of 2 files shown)


SergeAstapov:
Build fails in beta with error
```
dummy/tests/integration/components/changeset-test.js: Assertion Failed: Cannot invoke the `changeset` helper because it was shadowed by a local variable (i.e. a block param) with the same name. Please rename the local variable to resolve the conflict. (L2:C14) 
```

That assertion was introduced in https://github.com/emberjs/ember.js/pull/17132.

Fixed in #329

SergeAstapov:
@snewcomer I see there is large work happens in #317 and #327 (which is awesome!).
Meanwhile can this be releases as patch version? Is there anything I can help to push this and #330 forward?

snewcomer:
Yeah of course :). Will review by early next week.  Need to figure out what to pull into 2.0 and what to fix directly!

snewcomer:
Published 1.6.0!  Thanks @SergeAstapov.  Will add note however, that many users may want to stay on 1.3.0.  It depends on their use cases.

PR #330, opened 2018-11-22
Fix deprecations related to Ember.merge and ember-native-dom-helpers

SergeAstapov:
Build is failing on master, fixed in #329

### DIFF
--- a/addon/utils/assign.js
+++ b/addon/utils/assign.js
@@ -1,13 +1,6 @@
 // @flow
 
-import { assign as EmberAssign } from '@ember/polyfills';
-import { merge } from '@ember/polyfills'
-
-const assign = EmberAssign || Object.assign || _assign;
-
-function _assign(origin, ...sources) {
-  return sources.reduce((acc, source) => merge(acc, source), merge({}, origin));
-}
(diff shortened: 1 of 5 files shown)


addon/utils/assign.js
```diff
@@ -1,13 +1,6 @@
 // @flow
 
-import { assign as EmberAssign } from '@ember/polyfills';
```
SergeAstapov: I might be missing some point, but `Ember.assign` [falls back](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/polyfills/lib/assign.ts#L52) to `Object.assign` if it's available.

addon/utils/assign.js
```diff
@@ -1,13 +1,6 @@
...
-import { merge } from '@ember/polyfills'
-
-const assign = EmberAssign || Object.assign || _assign;
```
SergeAstapov: I might be missing some point, but `Ember.assign` [falls back](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/polyfills/lib/assign.ts#L52) to `Object.assign` if it's available.

PR #333, opened 2018-12-16
Belongs-To set to null or undefined

### LINKED ISSUE #199: Saving a relationship-model doesn't actually trigger the API call
Let's say I have a Post model with an author relationship and I want to create an ember-changeset for the author only, I would do this on the controller/component:

```
authorChangeset: computed('post', function () {
  let author = this.get('post.author');
  return new Changeset(author);
})
```

but when I try to persist the author, calling `save` on the changeset doesn't trigger the API call.
looking at the source code I think I found the issue:

https://github.com/DockYard/ember-changeset/blob/master/addon/index.js#L192
```
if (typeOf(content.save) === 'function') {
  savePromise = content.save(options);
}
```

I noticed that `post.save` is a function, but `author.save` is undefined. *but* `author.get('save')` would actually work. so because it's undefined, the `savePromise` remains `resolve(this)` and does nothing.

so my question would be, is there a reason why not to change it to:  
`if (typeOf(get(content, "save")) === 'function') {`

(or is this not supposed to work anyways? bad pattern?)

---

update 1: actually might be a duplicate of #127 - but that's already fixed/closed and I'm on 1.3.0, so I shouldn't have that issue. (anyways, when creating the changeset the relationship is already resolved)

snewcomer:
Fix relationship changeset save

Ref #315 
Close #199

### DIFF
--- a/tests/integration/main-test.js
+++ b/tests/integration/main-test.js
@@ -58,6 +58,14 @@ module('Integration | main', function(hooks) {
 
     assert.equal(user.get('profile.firstName'), 'Terry');
     assert.equal(user.get('profile.lastName'), 'Bubblewinkles');
+
+    changeset.set('profile', null);
+    assert.equal(changeset.get('profile'), null);
+
+    changeset.execute();
+
+    assert.equal(changeset.get('profile'), null);
+    assert.equal(user.get('profile').firstName, null);
   });
 
   test('it works for hasMany / firstObject', function(assert) {
--- a/tests/integration/main-test.js
+++ b/tests/integration/main-test.js
@@ -3,7 +3,7 @@ import { setupTest } from 'ember-qunit';
 import { run } from '@ember/runloop';
 import Changeset from 'ember-changeset';
 
-module('Integration | main', function(hooks) {
+module('scott Integration | main', function(hooks) {
   setupTest(hooks);
 
   hooks.beforeEach(function() {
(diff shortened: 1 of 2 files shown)


tests/integration/main-test.js
```diff
@@ -58,6 +58,16 @@ module('Integration | main', function(hooks) {
...
+
+    changeset.set('profile', null);
+    assert.equal(changeset.get('profile'), null);
+
+    changeset.execute();
+
+    assert.equal(changeset.get('profile'), null);
+    assert.equal(user.get('profile').firstName, null);
+    // failing test
+    // assert.equal(user.get('profile'), null);
```
snewcomer: @nickshot Is this the normal behaviour up until you `save()`?

tests/integration/main-test.js
```diff
@@ -58,6 +58,57 @@ module('Integration | main', function(hooks) {
...
     assert.equal(user.get('profile.lastName'), 'Bubblewinkles');
+
+    changeset.set('profile', null);
+    assert.equal(changeset.get('profile'), null, 'changeset profile is null');
+
+    changeset.execute();
+
+    assert.equal(changeset.get('profile'), null, 'changeset profile relationship is still null');
+    assert.equal(user.get('profile').get('firstName'), null, 'underlying user profile firstName is null');
+    assert.ok(user.get('profile'), 'user has yet to call save so still present as proxy');
```
snewcomer: @nickshot  Lmk if this is what you expect based on the tests.  I believe we could close #315 but wanted to check.

PR #335, opened 2019-01-03
Nested setters and getters improvement

snewcomer:
This PR will refactor methods around `validate/1` method and ensure `set` adds nested keys to change and errors maps.


snewcomer:
This mostly fixes the nested setter case.  Except for `set(changeset`.  Proxies are meant to provide traps for behaviour performed on it.  However, this case does not allow us to catch the behaviour.  This is particularly important when operating in template land as `(mut obj.person.firstName) value="target.value"` will implicitly do `set(`, circumventing our hooks.  I'm not sure of a way around at this point.

PR #351, opened 2019-04-26
Add changeset-get helper

### LINKED ISSUE #342: Deeply nested properties do not trigger template getter on rollback
## Version
2.0.0

## Test Case
https://github.com/neighborly/changeset-test

## Steps to reproduce
We are building a generalized `form` component based on `ember-changeset` that yields a set of input fields that are automatically bound to the changeset by a `fieldName` property. The `value` of these inputs is bound by a template getter to the `this.fieldName` property within the changeset, and `onInput` is bound to `action (changeset-set this.changeSet this.fieldName) value="target.value"`. The changeset picks-up on the changes to the input; however, when the form is reset, the changes clear from the changeset, but the inputs do not reset their values. 

It seems like the `{{get}}` helper only binds to the first token in a complex property key. This does *not* happen on properties that are at the top-level of the changeset model. For example, this will occur on a `name.first` property, but not on a `email` property.

Here is a demo app that illustrates the idea: https://github.com/neighborly/changeset-test

In this demo app change both of the values in the inputs, and click the reset button. You will only see the email field reset.

## Expected Behavior
Expected behavior is that any property would update when reset.

## Actual Behavior
Actual behavior is that the deeply nested properties do not update when reset.

josemarluedke:
This adds a new helper: `changeset-get` that allow us to get a value from template using dynamic field name with nested path. 

This closes #342 as it adds a way to go around the problem.

A working example of this helper can be found here: https://github.com/neighborly/changeset-test/pull/1

@jasonmevans Do you mind finishing up this PR?

 - [x] Add tests.
 - [x] Add documentation for the changes we made.

### DIFF
--- a/addon/helpers/changeset-get.ts
+++ b/addon/helpers/changeset-get.ts
@@ -1,9 +1,9 @@
-import Helper from "@ember/component/helper";
-import { observer } from "@ember/object";
-import { ChangesetDef } from "ember-changeset/types";
+import Helper from '@ember/component/helper';
+import { observer } from '@ember/object';
+import { ChangesetDef } from 'ember-changeset/types';
 
-const CONTENT = "_content";
-const CHANGES = "_changes";
+const CONTENT = '_content';
+const CHANGES = '_changes';
 
 export default class ChangesetGet extends Helper.extend({
   invalidate: observer(`changeset.${CONTENT}`, `changeset.${CHANGES}`, function(
@@ -17,12 +17,12 @@ export default class ChangesetGet extends Helper.extend({
   init() {
     // @ts-ignore
     super.init(...arguments);
-    this.set("changeset", null);
+    this.set('changeset', null);
   }
 
   compute(this: ChangesetGet, [changeset, fieldPath]: [ChangesetDef, string]) {
     if (this.changeset === null) {
-      this.set("changeset", changeset);
+      this.set('changeset', changeset);
     }
 
     if (!this.changeset) {
--- a/addon/helpers/changeset-get.ts
+++ b/addon/helpers/changeset-get.ts
@@ -1,0 +1,33 @@
+import Helper from '@ember/component/helper';
+import { observer } from '@ember/object'
+import { ChangesetDef } from 'ember-changeset/types';
+
+const CONTENT = '_content';
+const CHANGES = '_changes';
+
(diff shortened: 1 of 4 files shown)


addon/helpers/changeset-get.ts
```diff
@@ -0,0 +1,34 @@
...
+    this.recompute();
+  })
+}) {
+  changeset: ChangesetDef | null = null;
+
+  init() {
+    // @ts-ignore
+    super.init(...arguments);
+    this.set('changeset', null);
+  }
```
snewcomer: do we need this init?

addon/helpers/changeset-get.ts
```diff
@@ -0,0 +1,34 @@
+import Helper from '@ember/component/helper';
+import { observer } from '@ember/object';
+import { ChangesetDef } from 'ember-changeset/types';
+
+const CONTENT = '_content';
+const CHANGES = '_changes';
+
+export default class ChangesetGet extends Helper.extend({
+  invalidate: observer(`changeset.${CONTENT}`, `changeset.${CHANGES}`, function(
```
snewcomer: Do you have any documentation on `invalidate`?  Just curious b/c I thought I have seen it as a function call before instead of an observer...

addon/helpers/changeset-get.ts
josemarluedke: We actually can drop this.

addon/helpers/changeset-get.ts
josemarluedke: The name `invalidate` can be called anything. The key thing here is to call the recompute function, which will re-evaluate the compute method and render the result of that. 

Here is the documentation for the recompute: https://api.emberjs.com/ember/3.9/classes/Helper/methods/recompute?anchor=recompute

addon/helpers/changeset-get.ts
josemarluedke: Dropped.

addon/helpers/changeset-get.ts
snewcomer: 👍.  Was making sure we weren't inadvertently overriding something

jasonmevans:
Thanks for your help, @snewcomer and @josemarluedke!

PR #355, opened 2019-06-04
Fix push errors to update error property pr

alexspeller:

I found that I was having trouble binding to `errors.<someField>.validation` in a template without using `setNestedProperty` - I added some tests and made sure that `pushErrors` uses `setNestedProperty` like `addError` does

### DIFF
--- a/addon/index.ts
+++ b/addon/index.ts
@@ -438,11 +438,12 @@ export function changeset(
 
       let v = existingError.validation;
       validation = [...v, ...newErrors];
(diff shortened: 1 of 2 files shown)


balinterdi:
@snewcomer Would it be possible to publish a new release? The fact that the addon uses TS source files makes it hard to use a GH branch as a dependency (unless I'm mistaken, you have to fork and compile the files to JS to be able to use it).

Thank you.

Issue #356, opened 2019-06-18
Nested properties not changing on new Changeset


marshy2201:
## Version
```
"ember-changeset": "2.1.2"
```

## Test Case
https://ember-twiddle.com/02df9b64f83f5f844a76d63dff4184d8?openFiles=controllers.application.js%2C

## Steps to reproduce
* Create a changeset of an object 
* Set the changeset to a new changeset on the fly 
* Increment changeset fields 
* Call .rollback() 

## Expected Behavior
* Clicking on selected button group changes the fields to the new changeset
* Increment nested fields (max & min) on button click
* Rollback fields on button click

## Actual Behavior
* The ``` "results.max" ``` & ``` "results.min" ``` field (using changeset-get helper) doesn't change when creating new changeset
*  When new changeset created on click, the incrementNumber() & rollback() action stops working

muziejus:
I think I'm having the same problem:

https://ember-twiddle.com/595344913f994e7624ef09f33194f7ca?openFiles=templates.components.entry-table.hbs%2Ctemplates.components.entry-table.hbs

here, "Name" is a first-order property, but "Page" and "Other" are nested inside a `props` object. If you change one of the "Name" values in the second table ("as changeset"), then only that value changes. But if you change the "Page" or "Other" value, then the model changes as well.

muziejus:
This is beyond bizarre, because in my app proper, "name" doesn't update at all when you type in it, but the nested properties *do* update and go up to the validator (there's no validator here).

muziejus:
in the twiddle, when I change:

```
{{input
	type="text"
  value=(get changeset column)
  oninput=(action (mut (get changeset column)) value="target.value")
}}
```

to:

```
{{input
	type="text"
  value=(changeset-get changeset column)
  oninput=(action (changeset-set changeset column) value="target.value")
}}
```

the twiddle crashes.

marshy2201:
I think your twiddle crashes because your using the changeset template helpers in your ```{{table-cell}}``` component on model.entries that hasn't been set as a new changeset. By commenting out this code it works.

```
{{#each model.entries as |entry|}}
  <tr>
    {{#each columns as |column|}}
      <td>{{table-cell changeset=entry column=column}}</td>
    {{/each}}
  </tr>
{{/each}}
```

muziejus:
Thanks! I forgot I had the dummy table set up above! Now it seems to be working appropriately, and I'm testing the changes on the real app.

[...]

Issue #324, opened 2018-10-19
RFC: Pt. 3 - Evaluate use of native Proxy


snewcomer:
This is just an evaluation step.  Not necessarily saying we should go with it.  Need to evaluate after removing Relay implementation.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

Downside is no IE support

Ref #317 https://github.com/poteto/ember-changeset/issues/323 https://github.com/poteto/ember-changeset/issues/322

poteto:
I investigated this in https://github.com/poteto/validated-proxy, but gave up on it as there wasn't an obvious way to handle async sets and return a promise. I may have missed something though.

snewcomer:
@poteto Great point. Do you feel like returning a promise from a setter is necessary?  Right now, `set` returns a `ValidationResult`.  It seems like it was deemed true.

Setter happens `sync`.  Validation may be `async`. However, I'm wondering about the case for returning a `ValidationResult` from `set`.  We could require the user to call `.validate()` to get a `Promise<ValidationResult>` back. 

I believe this would allow us to use Proxy and probably squash some persistent issues around our current implementation.  

Would love to hear other's thoughts as well!

Issue #302, opened 2018-07-07
Recursive issue with applying changes


cah-briangantzler:
## Version
1.4.1 as required by ember-changeset-cp-validations

## Steps to reproduce
one of the values changed is an ember data model i.e. a belongs-to on a model

## Expected Behavior
not an error

## Actual Behavior
When applying the changes to the underlying model, there is a function that gets called _allKeys that is recursively looking for all the keys on the changes object. Since that contains a ember data model, there is a property store, which has a property recordArrayManager which has a property store.... you get a maximum calls exceeded.

I have played with the resolutions section of the package.json to use the 1.5.0-beta.0 version. This version does not include the _allKeys method. While I have got it working, the end result is that the most current version 1.4.1 is broken and will NOT work if any of the changed values are an ember data model.

Would request that at least the code that applies changes to the content object be back ported and release a 1.4.2 version

Thanks

snewcomer:
@cah-briangantzler Do you think we need to udpate ember-changeset-cp-validations?  

Working on getting a 1.5.0 release out, but might not be soon.  Recently fixed setting relationships on changesets in master so I'd be happy to work on that addon as well to make sure they are compatible.

cah-briangantzler:
Im looking at the code, and the only place _allKeys is being used is in the execute method. In 1.4.1 Im not sure what the reset properties is doing but it looks like you could just replace the 1.4.1 execute with the master execute and that should solve the problem for a 1.4.2 release before getting the 1.5.0 finished

snewcomer:
I'm going to wishfully assume the new 3.0.0 series fixes this!  Lmk if that is not the case.

PR #379, opened 2019-11-19
[MAJOR]: rewrite for v3

### LINKED ISSUE #245: RFC: Use es6 proxy
Didn't have time to write up an actual RFC. Here's a quick [proof of concept](https://ember-twiddle.com/44e1aac8ed7ad53cd45c7419c655e77d?openFiles=controllers.application.js%2C) where we are just naively trapping `proxy.set` to validate prior to setting the value on the target object.

If we go down this route, we should implement that as a separate JS library and then have this addon consume that.

snewcomer:
## Motivation
This major change attempts to solve a variety of (minor?) problems with `ember-changeset`'s more complicated use cases.  Notably, working with nested keys and chaining arbitrary nested keys and propagating changes to the UI layer (see this [test](https://github.com/poteto/ember-changeset/blob/4bb5fef85669b9549402afc7cd1a66ff1aa29934/tests/integration/components/changeset-test.js#L273) that was fixed as a result of these changes).

As a result, this necessarily **relies on Ember +3.13.** with `@tracked`.  No existing top level tests were harmed in this major rewrite.  Unused utils + tests were deleted && new utils + tests were added.

## Changes
1. On major pattern changed in this PR is instead of storing keys with arbitrary paths (`person.firstName`: value) internally on changes/errors, we store it as an object (`{ person: { firstName: value } }`) without dot notation.  This simplifies iterating, chaining and morphing objects and better aligns us with how JS objects work.

2. Moreover, we don't rely on `EmberObject` anymore.  Certain methods like `notifyPropertyChange` are imported and used where previously applied.

3. Because we are using a Proxy, async sets are not possible.  We currently rely on this for returning a maybe promise `ValidationResult`.  However, I'm not sure this was ever a necessary API.

4. We probably shouldn't rely on `Ember.set(changeset, ...)` anymore.  It will set on underlying model if the key is nested.  Recommend using `changeset.set(...)` or for single level keys with native setters, `changeset.foo = { key: 'mmbop' }`;

Will first release a beta release for ppl to try if and only if this PR passes the muster as compared to v2.

close #245 #324 #314 #357 
ref #356 #302


addon/-private/key-values.ts
```diff
@@ -0,0 +1,29 @@
+import isObject from '../utils/is-object';
+
+// 'foo.bar.zaz'
+let keysUpToValue: any[] = [];
```
Serabe: If this is used as an accumulator for recursion, this should be an argument to the function. Having it persisted across function calls can be error prone. If you don't want to expose the accumulator outside this module, you can use this common pattern:

```js
const myRecursiveFunctionWithAccumulator = (a, acc) => {
  /* do whatever recursive stuff you need to do */
};
export const myFunction = (a) => myRecursiveFunctionWithAccumulator(a, []);
```

addon/-private/merge-deep.ts
```diff
@@ -0,0 +1,132 @@
...
+	} catch(_) {
+		return false;
+	}
+}
+
+// Protects from prototype poisoning and unexpected merging up the prototype chain.
+function propertyIsUnsafe(target: any, key: string): Boolean {
+  return propertyIsOnObject(target, key) // Properties are safe to merge if they don't exist in the target yet,
+		&& !(Object.hasOwnProperty.call(target, key) // unsafe if they exist up the prototype chain,
+    && Object.propertyIsEnumerable.call(target, key)); // and also unsafe if they're nonenumerable.
```
Serabe: Correct indentation here. You might want to add prettier to keep consistency between tabs / spaces.

addon/-private/merge-deep.ts
```diff
@@ -0,0 +1,132 @@
...
+      possibleKeys.push(key);
+      buildPathToValue(possible, options);
+    } else {
+      possibleKeys = [];
+    }
+  });
+}
+
+/**
+ * `source` will always have a leaf key `value` with the proeprty we want to set
```
Serabe: s/proeprty/property/

addon/-private/merge-deep.ts
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * goal is to mutate target with source's properties, ensuring we dont encounter
+ * pitfalls of { ..., ... } spread syntax overwriting keys on objects that we merged
+ *
+ * @method mergeDeep
+ * @param target
+ * @param source
+ */
+export default function mergeDeep(target: any, source: any, options: Options = { safeGet: undefined, safeSet: undefined }): object | [any] {
+	options['safeGet'] = options.safeGet || function(obj: any, key: string): any { return obj[key] };
```
Serabe: Why the bracket?

addon/-private/normalize-object.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * traverse through target and unset `value` from leaf key
+ * Shallow copy here is fine because we are swapping out the leaf nested object
+ * rather than mutating a property in something with reference
+ *
+ * @method normalizeObject
+ * @param target
+ */
+export default function normalizeObject<T extends { [key: string]: any}>(target: T): T {
```
Serabe: Why the generic?

addon/-private/normalize-object.ts
```diff
@@ -0,0 +1,29 @@
...
+  let obj = { ...target };
+
+  for (let key in obj) {
+    if (key === 'value') {
```
Serabe: Can we get this check to be the first thing to do?

```js
if ('value' in target) { // Or hasOwnProperty
  return target.value;
}
```

That way we prevent creating a new object and traversing any property.

addon/-private/notifier.ts
```diff
@@ -0,0 +1,30 @@
...
+    for (let i = 0; i < this.listeners.length; i++) {
+      if (this.listeners[i] === callback) {
+        this.listeners.splice(i, 1);
+        return;
+      }
+    }
+  }
+
+  trigger(...args: any[]) {
+    this.listeners.slice(0).forEach(callback => callback(...args));
```
Serabe: Why do you need the `slice`?

addon/-private/notifier.ts
```diff
@@ -0,0 +1,30 @@
+// this statefull class holds and notifies
+import { INotifier } from '../types/evented';
+
+export default class Notifier implements INotifier {
```
Serabe: This is a generic version of Notifier that check that callbacks admit the right arguments.

http://www.typescriptlang.org/play/?ssl=26&ssc=22&pln=26&pc=27#code/C4TwDgpgBAwghgGwQIzgYwNYB4AqUIAewEAdgCYDOUcJIA2gLoB8UAvFABQB0PcATgHMKALig4AlGxYA3APYBLMgG4AUCsJhZfYFDIQAZnACuCHWgRwKVAHKzg8-fIh9c+IqUrVajJgG8VUIFQCPIUxCTOIpzcvIJRElJQcorijGpBUGiyJGF8RmjAWhyS-hkZwAAWoVwhYaSRbFCMqhkAvulBcGRkADKh4c4caIgo6BiiMVz8QqIJrDIKZCUBZYGV1bUDfBRcYEYUFUMjqJjiLatQfBDARnwknJLzUOs7VwC2stIQfXURfEdIE4YM4rQLtUGXCAfL4-LYA0aYCY8KZxWaPBYpKClC4vGr9erbVQQjL6LScBDXKDyRoABiUVKgWGeVR2mwJrNIAkq9PkAGpecsLmUHJxcWy-hQ6PIGGxWOxhoCxoKhTiWXjfpEuBQwCE0BAOPIADRQACMIJVFyuNzu5wt7SF9qC4PKfHkAgEg2R03iytWYvxEq1uv1NPEXFJfAAouhDgqERhEnGgZNveJzU6VK0gA

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+const RUNNING_VALIDATIONS = '_runningValidations';
+const BEFORE_VALIDATION_EVENT = 'beforeValidation';
+const AFTER_VALIDATION_EVENT = 'afterValidation';
+const AFTER_ROLLBACK_EVENT = 'afterRollback';
+const defaultValidatorFn = () => true;
+const defaultOptions = { skipValidate: false };
+
+export class BufferedChangeset implements IChangeset {
+  constructor(
+    obj: object,
```
Serabe: `public obj: object,`

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    validateFn: ValidatorAction = defaultValidatorFn,
```
Serabe: `public validateFn: ValidatorAction = defaultValidatorFn,`

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    validationMap: ValidatorMap = {},
```
Serabe: `public validationMap: ValidatorMap = {},`

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    options: Config = {}
+  ) {
+    this.obj = obj;
```
Serabe: Remove

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    this.validateFn = validateFn;
```
Serabe: Remove

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    this.validationMap = validationMap;
```
Serabe: Remove

addon/-private/notifier.ts
Serabe: In general, types like `Function`, `String`, `Number`, etc. should not be used.

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+  get isDirty() {
+    return !this.isPristine;
+  }
+
+  /**
+   * Stores change on the changeset.
+   *
+   * @method setUnknownProperty
+   */
+  setUnknownProperty<T> (
```
Serabe: Why the generic?

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    return this;
+  }
+
+  /**
+   * Executes the changeset and saves the underlying content.
+   *
+   * @method save
+   * @param {Object} options optional object to pass to content save method
+   */
+  save(
```
Serabe: `async` would be easier to read.

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+  addError<T> (
+    key: string,
+    error: IErr<T> | ValidationErr
+  ): IErr<T> | ValidationErr {
+    // Construct new `Err` instance.
+    let newError;
+    if (isObject(error) && !Array.isArray(error)) {
+      assert('Error must have value.', error.hasOwnProperty('value'));
+      assert('Error must have validation.', error.hasOwnProperty('validation'));
+      newError = new Err((<IErr<T>>error).value, (<IErr<T>>error).validation);
```
Serabe: Use `error as IErr<T>`. It is the preferred syntax as it is better supported by editors (specially when a plugin for TSX is active).

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    return Promise.resolve(this._validateKey(<string>key, this._valueFor(<string>key)));
+  }
+
+  /**
+   * Manually add an error to the changeset. If there is an existing
+   * error or change for `key`, it will be overwritten.
+   *
+   * @method addError
+   */
+  addError<T> (
```
Serabe: Given we are returning the same error we are given, we can do the following:

```ts
addError<T>(key: string, error: IErr<T>): IErr<T>
addError(key: string, error: ValidationErr): ValidationErr
addError<T>(key: string, error: IErr<T> | ValidationError) {
```

And we keep the type knowledge.

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    key: string,
+    error: IErr<T> | ValidationErr
+  ): IErr<T> | ValidationErr {
+    // Construct new `Err` instance.
+    let newError;
+    if (isObject(error) && !Array.isArray(error)) {
```
Serabe: This looks like a type guard. You can do:

```ts
const isIErr = <T>(error: unknown): error is IErr<T> => isObject(error) && !Array.isArray(error);
```

And you won't need to do a cast later in the branch.

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+    return this;
+  }
+
+  /**
+   * Checks to see if async validator for a given key has not resolved.
+   * If no key is provided it will check to see if any async validator is running.
+   *
+   * @method isValidating
+   */
+  isValidating(key: string | void): boolean {
```
Serabe: isValidating(key?: string): boolean {
}

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+  /**
+   * Validates a specific key
+   *
+   * @method _validateKey
+   * @private
+   */
+  _validateKey<T> (
+    key: string,
+    value: T
+  ): Promise<ValidationResult | T | IErr<T>> | T | IErr<T> | ValidationResult {
```
Serabe: I think this is a great candidate to an async function, so you don't need to split whether the validation is sync or async.

addon/-private/validated-changeset.ts
```diff
@@ -0,0 +1,933 @@
...
+   * runs the validator with the key and value
+   *
+   * @method _validate
+   * @private
+   */
+  _validate(
+    key: string,
+    newValue: unknown,
+    oldValue: unknown
+  ): ValidationResult | Promise<ValidationResult> {
```
Serabe: Always async.

addon/-private/set-deep.ts
```diff
@@ -0,0 +1,68 @@
...
+    target[path] = value;
+    return target;
+  }
+
+  for (let i = 0; i < keys.length; i++) {
+    let prop = keys[i];
+
+    if (!isObject(target[prop])) {
+      target[prop] = {};
+    } else if (isObject(target[prop]) && target[prop] instanceof Change) {
```
makepanic: you could extract `isObject(target[prop])` as variable to prevent checking it twice

addon/-private/set-deep.ts
```diff
@@ -0,0 +1,68 @@
...
+      // we don't want to merge new changes with a Change instance higher up in the obj tree
+      target[prop] = {};
+    }
+
+    // last iteration
+    if (i === keys.length - 1) {
+      result(target, prop, value);
```
makepanic: I think this could be refactored to extract the last iteration check out of the loop.
E.g. loop until `keys.length - 1` and always do the result call afterwards.

tsconfig.json
```diff
@@ -19,6 +20,7 @@
     "inlineSources": true,
     "baseUrl": ".",
     "module": "es6",
+    "skipLibCheck": true,
```
snewcomer: I hate to do this, but having a tough time with core libs and only including `@glimmer/tracking`

PieterJanVdb:
Mostly a question to inform myself as I don't know much about native Proxies. Does using Proxy instead of EmberObject for changesets now mean that we can use native setters (user.numbers.cell instead of users.set('numbers.cell')) or is it still recommended you use .set() for everything, as you mentioned in the 4th bullet point of Changes.

From what I can see in the tests you've added some native setters here and there but I guess it's not really recommended we do that everywhere with v3?

snewcomer:
@PieterJanVdb You are right!  Single level native setters work.  We would have to return a Proxy from the `get` trap to handle to handle nested key setters.

https://github.com/poteto/ember-changeset/pull/379/commits/54c6ba810f1224c0a59fc71d6055f8100d34769b

sandstrom:
@snewcomer Awesome! Does this change how relationships are handled?

More specifically, can v3 handle things such as this:

- Create a model `m`, e.g. an instance of `Mother` (Ember Data instance)
- Attach two children `Child` to the mother
- Rollback and have the children removed

snewcomer:
@sandstrom 👋 I have a test for rolling back a `hasMany` relationship.  Let me know if this does not work though.  With these new changes, it shouldn't be too hard to figure out if it doesn't work.

sandstrom:
@snewcomer Awesome, will take a look!

PR #380, opened 2019-11-28
Bring back skipped tests

snewcomer:
ref #379 

@sandstrom - yes but only with `changeset.set('dogs', [...oldDogs, newDog])`

### DIFF
--- a/tests/integration/components/changeset-test.js
+++ b/tests/integration/components/changeset-test.js
@@ -1,4 +1,4 @@
-import { module, test, skip } from 'qunit';
+import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { typeOf, isPresent } from '@ember/utils';
 import Changeset from 'ember-changeset';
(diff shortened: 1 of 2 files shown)


sandstrom:
@snewcomer Awesome news, will look into it this afternoon! 😄

rwwagner90:
@snewcomer @sandstrom is this saying you cannot do `addObject` or `pushObject` and must instead create a new array and set it on the changeset? We should document `hasMany` and `belongsTo` in the README probably?

sandstrom:
@rwwagner90 That's how I understood it yes, but I'm not 100% sure

snewcomer:
Some doc updates would probably be a good idea!  This is mainly due to our use of tracked (without track-built-ins)

rwwagner90:
@snewcomer thanks for the clarification. Is this only for hasMany or is this any array? I can take a stab at writing some docs.

snewcomer:
https://github.com/poteto/ember-changeset/blob/master/tests/integration/main-test.js#L187

Lmk if these tests are helpful at all and/or is different than what you though it should be!  And docs are always definitely appreciated!

PR #391, opened 2019-12-27
validated-changeset - exit beta

snewcomer:
https://github.com/validated-changeset/validated-changeset

This PR replaces much of the functionality present in this package with and plain TS library.  The goal of this extraction was to use in other contexts and frameworks.  We hope that we can obtain greater adoption of these amazing patterns and ideas here.  As an aside, we also have better type checks in the underlying pkg.

There are a few spots that are EmberData specific in which inheritance took over from `validated-changeset`.  Specifically `merge-deep` and the proxied `get`.  We will continue to look for more generic solutions once we are out of beta.

@josemarluedke @andreyfel 

A 3.0.0 release will follow and we will move out of beta!!


poteto:
@snewcomer I'm not comfortable with this change. You are essentially copying all the code from this repo and creating a new version without any attribution to the original author / contributors. You could have discussed this change with me first. Feel free to DM me on Twitter if you want to chat.

snewcomer:
As discussed offline, this extracted library started initially with the [major rewrite](https://github.com/poteto/ember-changeset/pull/379) we did a few months back to move us off EmberObject to allow its eventual use outside of Ember.  The hope is other ecosystems can take advantage of this phenomenal library.  As a result, we have moved this base library to it's proper place - https://github.com/validated-changeset/validated-changeset!!

Will update pkg once we can publish a `0.1.0` release!

PR #392, opened 2020-01-01
Node 10

snewcomer:


package.json
```diff
@@ -78,7 +78,7 @@
     "typescript": "^3.7.2"
   },
   "engines": {
-    "node": "8.* || >= 10.*"
+    "node": "10.* || 11.* || >= 12.*"
```
SergeAstapov: @snewcomer should this be just `10.* || >= 12.*`? Per https://github.com/nodejs/Release Node.js 11.x is not active release any more

snewcomer: Ah you are right.  I saw a small grey bar next to 11 in the maintenance schedule but not an active release as you said.  I'll fix!

PR #393, opened 2020-01-09
Fix TS type for changeset.errors

josemarluedke:
Errors have the "validation" key, which has the error messages from validations.

### DIFF
--- a/addon/-private/validated-changeset.ts
+++ b/addon/-private/validated-changeset.ts
@@ -37,7 +37,7 @@ import {
   ValidatorMap
 } from 'ember-changeset/types';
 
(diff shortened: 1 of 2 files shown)


addon/-private/validated-changeset.ts
```diff
@@ -157,13 +157,9 @@ export class BufferedChangeset implements IChangeset {
     }
 
     return keys(obj).map(key => {
-      let value = transform(obj[key]);
+      let { value, validation } = transform(obj[key]);
 
-      if (isObject(value)) {
```
josemarluedke: This was always true as it as always returned an object from the transform method.

PR #395, opened 2020-01-10
Add safeSet for Ember tracking context

### LINKED ISSUE #394: [3.0.0-beta.4] mutating the model in Glimmer Components
I'm using the latest beta version in a glimmer component and I have hit a wall about changeset mutating the underlying model.

```
Error: Assertion Failed: You attempted to update [object Object].email to "something-error", but it is being tracked by a tracking context, such as a template, computed property, or observer. In order to make sure the context updates properly, you must invalidate the property when updating it. You can mark the property as `@tracked`, or use `@ember/object#set` to do this.
    at assert (index.js:172)
```

I have created a reproduction app, see the link below. Note that I added two examples, one with the model being tracked and another with the model as a class with a specific field as tracked. 

## Version
3.0.0-beta.4

## Test Case
https://github.com/josemarluedke/--changeset-glimmer 

## Steps to reproduce
Close repo, run app, change input value and submit. See an error in console.

snewcomer:
close #394 

@josemarluedke 👋 Thank you for the issue!  I believe this is because we need to use `Ember.set`.  This PR makes it work.

One question is - what was the reason to track the model on the Component instance in your example?  At first I assumed this wasn't necessary.  Let me know your thoughts!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,8 +51,8 @@ jobs:
       env: EMBER_TRY_SCENARIO=ember-3.13
     - env: EMBER_TRY_SCENARIO=ember-release
     - env: EMBER_TRY_SCENARIO=ember-beta
-    - env: EMBER_TRY_SCENARIO=ember-canary
     - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
+    - env: EMBER_TRY_SCENARIO=ember-canary
 
 before_install:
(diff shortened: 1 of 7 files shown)


josemarluedke:
@snewcomer Thanks for working on this. I don't think the tracked is going to be required as I had in the example. I used it as a way to see if would help with anything.

PR #399, opened 2020-01-11
 Collect errors correctly

### LINKED ISSUE #396: Errors from nested keys have incorrect key
Doing validations on model that has nested keys results in having a wrong error object (`changeset.errors`). 

## Version
3.0.0-beta.6

## Test Case
https://github.com/josemarluedke/--changeset-glimmer (Commit: https://github.com/josemarluedke/--changeset-glimmer/commit/171cda34825d4025f0fd136eeafac40e27a715af)

## Steps to reproduce
Close repo, run app, submit the form. See the errors array on the screen.

## Expected Behavior

Expected errors array should be something like this:
```json
[
  {
    "key": "user.firstName",
    "value": "",
    "validation": [
      "the validation message here"
    ]
  }
]
```

## Actual Behavior

```json
[
  {
    "key": "user"
  }
]
```

snewcomer:
close #396

### DIFF
--- a/addon/types/index.ts
+++ b/addon/types/index.ts
@@ -70,8 +70,8 @@ export interface ChangesetDef {
   _runningValidations: RunningValidations,
   _bareChanges: { [s: string]: any },
 
-  changes: any, // { key: string; value: any; }[], //ComputedProperty<object[], object[]>,
-  errors: any, // { key: string; value: any; validation: ValidationErr | ValidationErr[] }[], //ComputedProperty<object[], object[]>,
+  changes: object[],
+  errors: object[],
   error: object,
   change: object,
   data: object,
--- a/addon/types/index.ts
+++ b/addon/types/index.ts
@@ -71,7 +71,7 @@ export interface ChangesetDef {
   _bareChanges: { [s: string]: any },
 
   changes: any, // { key: string; value: any; }[], //ComputedProperty<object[], object[]>,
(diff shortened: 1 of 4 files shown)


addon/types/index.ts
```diff
@@ -70,8 +70,8 @@ export interface ChangesetDef {
   _runningValidations: RunningValidations,
   _bareChanges: { [s: string]: any },
 
-  changes: any, // { key: string; value: any; }[], //ComputedProperty<object[], object[]>,
-  errors: { key: string; value: any; validation: ValidationErr | ValidationErr[] }[], //ComputedProperty<object[], object[]>,
```
josemarluedke: @snewcomer Any specific reason for this change? It was super helpful to have the actual shape of the error here.  I would suggest making a type (or an interface) specifically for the error. 

Here is an example of a function that is written in TS which was working fine before this change:

```ts
  get errors(): string[] {
    const fieldErrors = this.args.changeset.errors.filter(error => {
      return error.key === this.fullFieldName;
    });

    return fieldErrors.reduce(
      (errors: string[], error: ValidationError): string[] => {
        return [...errors, ...error.validation];
      },
      []
    );
  }
```

Now typescript is not happy, with the following error: 

```
Property 'key' does not exist on type 'object'.

25       return error.key === this.fullFieldName;
```

Probably the same would apply for `changes` above.

snewcomer: This should be an easy fix...

Issue #398, opened 2020-01-11
Multiple properties in nested key loses data on changeset.set


josemarluedke:
Weird behavior when a model has more than one property inside of a nested key. It seems that the data is lost in the changeset for a moment. 

In the example, we have the following model: 

```js
  model = {
    email: 'hi@gmail.com',
    user: {
      firstName: null,
      lastName: null
    },
    notifications: {
      email: true,
      sms: true
    }
  }
```

It's probably easier to see the issue in the gifs below.

## Version
3.0.0-beta.7

## Test Case
https://github.com/josemarluedke/--changeset-glimmer (Commit: https://github.com/josemarluedke/--changeset-glimmer/commit/10b7be96ea362378c0926b08cb504e705a1c4316)

## Steps to reproduce
Close repo, run app, change "user.firstName", change "user.lastName", Sumit form, change "user.firstName" again. You will see "user.lastName" will lose it's data.
See gifs below:

### Case 1, changing one checkbox, makes the second checkbox lose its data. 
![Screen Recording 2020-01-10 at 09 45 PM](https://user-images.githubusercontent.com/230476/72199642-a82fdd00-33f3-11ea-932d-f8aa5d05f658.gif)

### Case 2, changing the value of input after the data is saved in the model makes the other input to lose its data.
![Screen Recording 2020-01-10 at 09 44 PM](https://user-images.githubusercontent.com/230476/72199644-a82fdd00-33f3-11ea-85bf-f64149b0c269.gif)

## Expected Behavior

Both properties in the same nested key should have its data intact.

@snewcomer Sorry for these issues, it's hard to explain some of them. Let me know if you have any questions.

snewcomer:
@josemarluedke Thank you thank you! Am I on the right path with the test I added in the below PR?  What am I missing from your reproduction?

https://github.com/validated-changeset/validated-changeset/pull/9/files

josemarluedke:
@snewcomer Commented in that PR. I suggested adding a new case where only one property is set at a time. The tests seem to pass there fine, so probably the issue is in the Ember integration. Although, it doesn't seem you are using that lib here yet.

snewcomer:
@josemarluedke Alright I had the chance to checkout your fork.  I think the issue was the README.  the `value` attribute on the input should reference the model.  B/c live `value` on the input doesn't reflect back the attribute, this would be recommended.  What do you think?

[...]

PR #401, opened 2020-01-11
Add set test

snewcomer:
https://github.com/validated-changeset/validated-changeset/pull/9#discussion_r365536512

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,8 +110,12 @@ Then, in your favorite form library, simply pass in the `changeset` in place of 
 ```hbs
 {{! dummy-form/template.hbs}}
 <form>
-  <Input @value={{changeset.firstName}} />
-  <Input @value={{changeset.lastName}} />
+  <input
+    value={{model.firstName}}
+    oninput={{action (changeset-set this.changeset "model.lastName") value="target.value"}}>
+  <input
+    value={{model.lastName}}
+    oninput={{action (changeset-set this.changeset "model.lastName") value="target.value"}}>
 
   <button {{on "click" this.submit changeset}}>Submit</button>
   <button {{on "click" this.rollback changeset}}>Cancel</button>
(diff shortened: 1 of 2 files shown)


tests/unit/changeset-test.js
```diff
@@ -685,6 +682,60 @@ module('Unit | Utility | changeset', function(hooks) {
...
+    assert.equal(dummyModel.org.usa.ny, 'nil');
+    assert.equal(dummyModel.org.usa.mn, 'MN');
+
+    c.save();
+
+    assert.equal(c.get('org.usa.ny'), 'nil2');
+    assert.equal(c.get('org.usa.mn'), 'undefined');
+    assert.equal(dummyModel.org.usa.ny, 'nil2');
+    assert.equal(dummyModel.org.usa.mn, 'undefined');
+  });
```
snewcomer: @josemarluedke Does this test look right to you?

josemarluedke: Yep. Looks good! Interesting that it is passing. Not sure how to capture that bug in a test then.

README.md
```diff
@@ -110,8 +110,12 @@ Then, in your favorite form library, simply pass in the `changeset` in place of
 ```hbs
 {{! dummy-form/template.hbs}}
 <form>
-  <Input @value={{changeset.firstName}} />
-  <Input @value={{changeset.lastName}} />
+  <input
+    value={{model.firstName}}
+    oninput={{action (changeset-set this.changeset "model.lastName") value="target.value"}}>
```
snewcomer: @josemarluedke Do these README updates look ok to you?  ie. the `model` as the `value` as noted https://github.com/poteto/ember-changeset/issues/398#issuecomment-573366958

josemarluedke: I don't think this change is correct. I believe forms should always reflect the in-flight model. This way, we can revert changes and have forms reflect it without any issues. Also, If you have a form in a tab layout for example, and you hide half of the form then go back to that, the form would show the incorrect data, the user would think they have lost their changes.

snewcomer: reverted.  Thanks a ton for the insight.

PR #404, opened 2020-01-14
Fix changeset get for nested properties with changes

### LINKED ISSUE #398: Multiple properties in nested key loses data on changeset.set

josemarluedke:
This attempts to fix the issue in #398 and adds a test that covers the problem identified in the issue.

Closes #398.

### DIFF
--- a/addon/-private/validated-changeset.ts
+++ b/addon/-private/validated-changeset.ts
@@ -49,6 +49,12 @@ const AFTER_VALIDATION_EVENT = 'afterValidation';
 const AFTER_ROLLBACK_EVENT = 'afterRollback';
 const defaultValidatorFn = () => true;
 const defaultOptions = { skipValidate: false };
+const isEmberDataObject = (obj: Object) => {
+  let keys = Object.keys(obj)
+  return keys.indexOf('content') !== -1 &&
+         keys.indexOf('isFulfilled') !== -1 &&
+         keys.indexOf('isRejected') !== -1
+}
 
 export class BufferedChangeset implements IChangeset {
   constructor(
@@ -851,13 +857,30 @@ export class BufferedChangeset implements IChangeset {
       }
 
       if (result && isObject(result)) {
-        return normalizeObject(result);
+        result = normalizeObject(result);
+        let content: Content = this[CONTENT];
+
+        // Merge the content with the changes to have a complete object for a nested property.
+        // Given a object with nested property and multiple properties inside of it, if the 
+        // requested key is the top most nested property and we have changes in of the properties, we need to 
+        // merge the original model data with the changes to have the complete object.
+        // eg. model = { user: { name: 'not changed', email: 'changed@prop.com'} }
+        if (isObject(content[baseKey]) && !isEmberDataObject(content[baseKey])) {
+          let data: Record<string, any> = {}
+          Object.keys(content[baseKey]).forEach(k => {
+            data[k] = this.getDeep(content, `${baseKey}.${k}`)
+          })
+          return Object.assign(data, result);
+        }
+
+        return result
       }
 
       if (result) {
         return result.value;
       }
     }
+
 
     // return getters/setters/methods on BufferedProxy instance
     if (this[key]) {
--- a/tests/dummy/app/components/changeset-form.ts
+++ b/tests/dummy/app/components/changeset-form.ts
@@ -5,6 +5,7 @@ import Changeset from 'ember-changeset'
 export default class ChangesetForm extends Component {
   model = {
     user: {
+      name: 'someone',
       email: 'something'
     },
     cid: '1'
--- a/tests/dummy/app/templates/components/changeset-form.hbs
+++ b/tests/dummy/app/templates/components/changeset-form.hbs
@@ -1,6 +1,8 @@
+<h2>Changeset Name: <span data-test-changeset-user-name>{{this.changeset.user.name}}</span></h2>
 <h2>Changeset Email: <span data-test-changeset-user-email>{{this.changeset.user.email}}</span></h2>
 <h2>Changeset cid: <span data-test-changeset-cid>{{this.changeset.cid}}</span></h2>
 <br />
+<h2>Model Email: <span data-test-model-user-name>{{this.model.user.name}}</span></h2>
 <h2>Model Email: <span data-test-model-user-email>{{this.model.user.email}}</span></h2>
 <h2>Model cid: <span data-test-model-cid>{{this.model.cid}}</span></h2>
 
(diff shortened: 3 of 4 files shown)


josemarluedke:
@snewcomer This PR is ready for review. 

Unfortunately, I had to add a workaround Ember Data Objects. I hope this PR makes sense. If you have a better approach to solve this issue, let me know.

addon/-private/validated-changeset.ts
```diff
@@ -49,6 +49,12 @@ const AFTER_VALIDATION_EVENT = 'afterValidation';
 const AFTER_ROLLBACK_EVENT = 'afterRollback';
 const defaultValidatorFn = () => true;
 const defaultOptions = { skipValidate: false };
+const isEmberDataObject = (obj: Object) => {
```
snewcomer: what do you think about this take to check if an ember-data model? #406

josemarluedke: I believe it doesn't do the same thing. I could be wrong, just testing out to see if all tests pass.

addon/-private/validated-changeset.ts
```diff
@@ -851,14 +857,27 @@ export class BufferedChangeset implements IChangeset {
...
+        result = normalizeObject(result);
+         let content: Content = this[CONTENT];
+
+        if (isObject(content[baseKey]) && !isEmberDataObject(content[baseKey])) {
+          let data: { [key: string]: any } = {}
+          Object.keys(content[baseKey]).forEach(k => {
+            data[k] = this.getDeep(content, `${baseKey}.${k}`)
+          })
+          return Object.assign(data, result);
+        }
```
snewcomer: Do we need this because Ember is asking for `user` and not the leaf key?  I was under the assumption that the key passed was the full path (but perhaps was pre Octane)...For example I see  the`key` arg as `cid` but never `user.name` for example.  Just `user` as the argument to `get`.

Overall this is great!  Can we add a comment on what we are trying to do?  What I gather is you are taking the `CHANGES` and merging with the `CONTENT` at that level in the object, correct?

josemarluedke: Yeah, I thought it did request the full path as well, but it doesn't. If I render `user.name`, the requested key is just `user`. 

We need this "merge" because the `user ` has changed, but only one property in it, causing the return value to be incomplete when having more than one property. 

The change here fetches the actual content and merge the changes on top of it, having complete data for a nested property.

I will add some comments to the code.

snewcomer: Screen Shot 2020-01-15 at 12 53 52 PM

Yeah looks like it traverses to the parent to find the value :( . Perhaps was a recent update, perhaps not.  

In any case, the fix you have here seems like the correct one.

addon/-private/validated-changeset.ts
```diff
@@ -851,14 +857,32 @@ export class BufferedChangeset implements IChangeset {
       }
 
       if (result && isObject(result)) {
-        return normalizeObject(result);
-      }
+        result = normalizeObject(result);
+         let content: Content = this[CONTENT];
```
snewcomer: minor spacing issues when you have a chance (should be caught by a linter and is fixed in the `validated-changeset` repo)

addon/-private/validated-changeset.ts
```diff
@@ -851,14 +857,32 @@ export class BufferedChangeset implements IChangeset {
...
+
+        // Merge the content with the changes to have a complete object for a nested property.
+        // Given a object with nested property and multiple properties inside of it, if the 
+        // requested key is the top most nested property and we have changes in of the properties, we need to 
+        // merge the original model data with the changes to have the complete object.
+        // eg. model = { user: { name: 'not changed', email: 'changed@prop.com'} }
+        if (isObject(content[baseKey]) && !isEmberDataObject(content[baseKey])) {
+          let data: { [key: string]: any } = {}
```
snewcomer: `let data: Record<string, any> = {}`

is another option

addon/-private/validated-changeset.ts
snewcomer: Did you see if that worked?  We had another issue on EmberData objects and the fix on #406 seemed to fix it for them...

addon/-private/validated-changeset.ts
josemarluedke: Changed.

addon/-private/validated-changeset.ts
josemarluedke: Yeah, I had to disable all prettier and formatting in my editor to be able to work in these files, otherwise, my changes would be a lot of formatting. 

Anyway, I pushed a fix for the spacing.

addon/-private/validated-changeset.ts
```diff
@@ -851,14 +857,32 @@ export class BufferedChangeset implements IChangeset {
...
+        if (isObject(content[baseKey]) && !isEmberDataObject(content[baseKey])) {
+          let data: Record<string, any> = {}
+          Object.keys(content[baseKey]).forEach(k => {
+            data[k] = this.getDeep(content, `${baseKey}.${k}`)
+          })
+          return Object.assign(data, result);
+        }
+
+          return result
+        }
```
snewcomer: These last two lines as well....

addon/-private/validated-changeset.ts
snewcomer: Also noting that either way is fine with me.  Whatever seems more canonical is what we should do across the repo.

josemarluedke: @snewcomer I just tried a similar approach and it didn't work. If there isn't any other way to see if an object is and Ember Data object, then I think this change is fine.

snewcomer: Ok.  I don't want to drag this PR on as you seem to have fixed quite a good bug that surely will help others!

Ultimately, we would want a `util` to determine if something is an Ember Data object.  If #406 isn't it, we probably need to make sure we have the canonical way of doing so.

@luketheobscure

Issue #403, opened 2020-01-13
Saving belongsTo() relationship doesn't work


andreyfel:
## Version
ember-changeset@^3.0.0-beta.9

## Test Case

Profile model:
```
class Profile extends Model {
  @attr('string') name;
}
```

User model:
```
class User extends Model {
  @attr('string') email;
  @belongsTo('profile') profile;
}
```

Creating a user with a profile and saving it:
```
const profiles = await this.store.finsAll('profile');
const user = this.store.createRecord('user');
const changeset = new Changeset(user, ...);
changeset.set('email', 'example@email.com');
changeset.set('profile', profile[0]);
changeset.save();
```

## Expected Behavior
User is saved

## Actual Behavior
An exception is thrown:
```
Uncaught (in promise) Error: Assertion Failed: Cannot call set with 'name' on an undefined object.
    at assert (index.js:128)
    at _set2 (index.js:2632)
    at setPath (index.js:2692)
    at Object._set2 [as safeSet] (index.js:2642)
    at merge-deep.js:92
    at Array.forEach (<anonymous>)
    at mergeTargetAndSource (merge-deep.js:81)
    at mergeDeep (merge-deep.js:150)
    at Proxy.execute (index.js:196)
    at Proxy._callee$ (validated-changeset.js:261)
```

andreyfel:
Changeset contains ember model in the profile field and apparently it tries to make a deep copy of it on save.

andreyfel:
I've modified one test to illustrate the issue in the testuite: #405 
```
Error: Assertion Failed: Cannot delegate set('firstName', Grace) to the 'content' property of object proxy <(unknown):ember183>: its 'content' is undefined.
    at assert (http://localhost:7357/assets/vendor.js:38210:15)
    at Proxy.setUnknownProperty (http://localhost:7357/assets/vendor.js:27561:48)
    at set (http://localhost:7357/assets/vendor.js:17273:11)
    at setPath (http://localhost:7357/assets/vendor.js:17301:14)
    at Object.set [as safeSet] (http://localhost:7357/assets/vendor.js:17248:14)
    at http://localhost:7357/assets/vendor.js:67256:23
    at Array.forEach (<anonymous>)
    at mergeTargetAndSource (http://localhost:7357/assets/vendor.js:67245:21)
    at mergeDeep (http://localhost:7357/assets/vendor.js:67312:14)
    at Proxy.execute (http://localhost:7357/assets/vendor.js:66796:48)
```

[...]

PR #405, opened 2020-01-14
User save test: create user without the profile

andreyfel:
Illustration of #403

### DIFF
--- a/tests/integration/main-test.js
+++ b/tests/integration/main-test.js
@@ -87,9 +87,10 @@ module('Integration | main', function(hooks) {
       assert.ok(true, 'user save was called')
     }
 
-    let user = this.store.createRecord(userType, { profile, save });
+    let user = this.store.createRecord(userType, { save });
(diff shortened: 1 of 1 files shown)


snewcomer:
Is this PR close-able?

andreyfel:
@snewcomer I rebased it and it is still failing. I'm wondering if it is supposed to be supported in this way?
```
  changeset.set('profile', profile);
  changeset.set('profile.firstName', 'Grace');
```

snewcomer:
Yeah it seems we can't do

```js
options.safeSet(Proxy, 'firstName', next.value);
```

I wonder why that is 🤔 There is probably something in the merge-deep util we need to look at...

snewcomer:
What I am seeing is that you effectively have to set the `profile` on the user initially.  Based on what I understand, this is inherent to ED and not ember-changeset.  

This is a minor inconvenience; however, not solvable.  

At minimum, we can document our findings.  What do you think?

PR #406, opened 2020-01-14
Failing test for ember data model.

luketheobscure:
Test case for #407.


addon/utils/merge-deep.ts
```diff
@@ -39,8 +39,18 @@ function propertyIsOnObject(object: any, property: any) {
   }
 }
 
+function hasEmberDataProperty(target: any, key: string, options: Options): Boolean {
```
luketheobscure: fwiw, this should have the lowercase `boolean` as the return type, but I tried to follow the conventions that were already here.

PR #408, opened 2020-01-17
Fix a bug with unsafe properties

andreyfel:
possibleKeys were shared between the branches during the traverse.
Each node could push to it making it polluted for the next branch.
Source like: {a: {}, b: {c: {}}} could produce paths like:
['a', 'a.b.c'] instead of ['a', 'b.c'];

### DIFF
--- a/tests/unit/utils/merge-deep-test.js
+++ b/tests/unit/utils/merge-deep-test.js
@@ -35,4 +35,32 @@ module('Unit | Utility | merge deep', () => {
 
     assert.deepEqual(value, { company: { employees: ['Jull', 'Olafur']} }, 'has right employees');
   });
+
+  test('it works with unsafe properties', async function(assert) {
+    class A {
+      _boo = 'bo';
+
+      get boo() {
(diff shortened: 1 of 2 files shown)


tests/unit/utils/merge-deep-test.js
```diff
@@ -35,4 +35,32 @@ module('Unit | Utility | merge deep', () => {
 
     assert.deepEqual(value, { company: { employees: ['Jull', 'Olafur']} }, 'has right employees');
   });
+
+  test('it works with unsafe properties', async function(assert) {
```
snewcomer: What do you mean by "unsafe" here?  Just curious b/c we also use this term to reference prototype poisoning in this codebase

andreyfel: I wasn't sure about the term, I used exactly the same term I found in the source code. I didn't find tests which were testing that branch of logic, so I've tried to "reverse engineer" the code to make tests go through that branch.

PR #409, opened 2020-01-17
Extend test coverage to sync relationships

andreyfel:
This PR extends test coverage for sync and async relationships.

### DIFF
--- a/tests/integration/main-test.js
+++ b/tests/integration/main-test.js
@@ -68,8 +68,7 @@ module('Integration | main', function(hooks) {
     changeset.execute();
 
     assert.equal(changeset.get('profile'), null, 'changeset profile relationship is still null');
-    assert.equal(user.get('profile').get('firstName'), null, 'underlying user profile firstName is null');
-    assert.ok(user.get('profile'), 'user has yet to call save so still present as proxy');
+    assert.equal(user.get('profile.firstName'), null, 'underlying user profile firstName is null');
   }
 
   test('it works for belongsTo', async function(assert) {
--- a/tests/integration/main-test.js
+++ b/tests/integration/main-test.js
@@ -209,4 +209,33 @@ module('Integration | main', function(hooks) {
   test('it can rollback sync hasMany', async function(assert) {
     await testRollbackHasMany.call(this, assert, 'sync-user');
   });
+
+  async function testInitiallyEmptyRelationships(assert, userType) {
+    let profile = this.store.createRecord('profile');
+    let user =  this.store.createRecord(userType);
+
+    let changeset = new Changeset(user);
+
+    changeset.set('profile', profile);
+    const dogs = [
+      this.store.createRecord('dog'),
+      this.store.createRecord('dog', { breed: 'Münsterländer' })
+    ];
+
+    changeset.set('dogs', dogs);
+
+    changeset.execute();
+
+    assert.equal(user.get('profile.firstName'), 'Bob', 'Profile is set on user');
+    assert.equal(user.get('dogs.firstObject.breed'), 'rough collie');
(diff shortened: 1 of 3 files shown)


snewcomer:
I like this!!  👍 Happen to know why the two tests are failing?

Also to confirm, there is no addon code changes needed right?

andreyfel:
After recent changes added in #404 those 2 tests started failing. It is caused by the fact that `isEmberDataObject` util method doesn't work with sync relationships. We need a more reliable approach.
cc: @josemarluedke

snewcomer:
Great points!  Yes you are right.  A couple of points.

1.  We need to add a check for `obj._internalModel`, in the EmberData check.  Seems that initial check is not comprehensive.
2.  We aren't checking if is a type of array in the referenced PR before doing Object.assign.  So it would have to look something like this...

`if (!Array.isArray(result) && isObject(content[baseKey]) && !isEmberDataObject(content[baseKey])) {`

snewcomer:
I merged the Array check #413

andreyfel:
_internalModel is not a public API, so, it can be removed at any moment. I've asked for advice in the ED discord channel.

andreyfel:
@runspired suggested that we should ask the user if the object is an Ember Data Model or not.
So, we should either have something like a flag passed in the Changeset constructor which says if an object is an ED model or even a separate class for ED models.

snewcomer:
@andreyfel I like the idea of explicitly saying it is an EmberData object.  However, this might get a little murky when an object's nested values are EmberData objects.

A few thoughts.

1.  We can bring back the `changeset-get` helper.  Ref #414.  Ultimately, requesting for the leaf key should actually ask for the leaf key and ideally the logic introduced in #404 shouldn't be necessary.
2.  However, based on `master`, it is necessary if the end user calls `changeset.get('user').lastName` when only `user.firstName` has changes because right now, `lastName` will be the `model` value.  I think there is a fundamental problem I have built into the proxied `get` that has led this discussion astray.  It is this line - https://github.com/poteto/ember-changeset/blob/master/addon/-private/validated-changeset.ts#L896.  `changeset.get('user').lastName` should return `undefined` and not the original value on the model.  

The combination of these two would alleviate all our problems, albeit, a stricter API.  Do you have any thoughts?

tests/integration/main-test.js
```diff
@@ -171,5 +200,42 @@ module('Integration | main', function(hooks) {
...
+      this.store.createRecord('dog', { breed: 'Münsterländer' })
+    ];
+
+    changeset.set('dogs', dogs);
+
+    changeset.execute();
+
+    assert.equal(user.get('profile.firstName'), 'Bob', 'Profile is set on user');
+    assert.equal(user.get('dogs.firstObject.breed'), 'rough collie');
+    assert.equal(user.get('dogs.lastObject.breed'), 'Münsterländer');
```
snewcomer: FYI I don't think these array accessors work (at the moment at least)

andreyfel: Well, they work for EmberArray and `dogs` is an EmberArray.

snewcomer: Oh you are right...we are passing in Ember's getter...

snewcomer:
@andreyfel Is this PR g2g?  I'm happy with it in its current state as long as the latest master is merged in...

andreyfel:
Yes, I've rebased it today. The pr can be merged and we can think of refactoring after that.

andreyfel:
@snewcomer yesterday I wrote a comment but it appears I forgot to publish it.
Regarding `object's nested values are EmberData objects`. Is this feature supported? Are there are tests for that scenario? Maybe we don't need to support that?
Thus PR is kind of a bandaid fix to make tests pass and merge them in but I believe we should do something about it in the long term.

Issue #410, opened 2020-01-17
No typescript type for errors


josemarluedke:
In a recent PR, TypeScript type was removed for errors, it's just an "object" now. 

See comment: https://github.com/poteto/ember-changeset/pull/399#discussion_r365538382

snewcomer:
lmk if beta-12 fixes it!

josemarluedke:
Fixed in #411. Thanks.

PR #411, opened 2020-01-17
PublicErrors type

snewcomer:
ref #410

### DIFF
--- a/addon/types/index.ts
+++ b/addon/types/index.ts
@@ -32,6 +32,8 @@ export type Errors<T> = {
   [s: string]: IErr<T>
 };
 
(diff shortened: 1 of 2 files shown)


addon/types/index.ts
```diff
@@ -70,8 +72,8 @@ export interface ChangesetDef {
   _runningValidations: RunningValidations,
   _bareChanges: { [s: string]: any },
 
-  changes: object[],
-  errors: object[],
+  changes: Record<string, any>[],
+  errors: PublicErrors,
```
snewcomer: @josemarluedke whaddy think?

PR #413, opened 2020-01-18
Follow up: Check array type as well

snewcomer:
ref #404


josemarluedke:
@snewcomer Thanks for catching and fixing this.

PR #414, opened 2020-01-19
Bring back changeset-get for nested getter

snewcomer:
Ideally our templating layer asks for the leaf key.  However this is not the case.  If looking for `this.user.name`, `user` will be propagated to the template layer.  

By requiring `changeset-get` for nested properties, we could remove the logic in #404.  Although still thinking if this is a good idea or not.


josemarluedke:
Just to clarify that this change does not remove the need from the changes in #404. I originally caught that issue by using `changeset.get`. (See https://github.com/josemarluedke/frontile/pull/24)

snewcomer:
I doesn't...yet.  Relying on it does create other problems though as noted in #409.

With the everpresent EmberData object arbitrarily nested at any level, we may indicate that users request up to the leaf key.  In templates, this helper makes it a necessity.  In `JS` land, we can suggest this.  Overall, I'm not exactly sure how to proceed though.  Will ping you if we decide to do anything with that though.

PR #415, opened 2020-01-20
Add failing test for nested properties with booleans

josemarluedke:
This adds a test for a bug I encountered today.

cc/ @snewcomer

### DIFF
--- a/tests/dummy/app/templates/components/changeset-form.hbs
+++ b/tests/dummy/app/templates/components/changeset-form.hbs
@@ -1,10 +1,14 @@
 <h2>Changeset cid: <span data-test-changeset-cid>{{this.changeset.cid}}</span></h2>
 <h2>Changeset Email: <span data-test-changeset-user-email>{{changeset-get this.changeset "user.email"}}</span></h2>
 <h2>Changeset Name: <span data-test-changeset-user-name>{{changeset-get this.changeset "user.name"}}</span></h2>
+<h2>Changeset notifications email: <span data-test-changeset-notifications-email>{{changeset-get this.changeset "notifications.email"}}</span></h2>
+<h2>Changeset notifications sms: <span data-test-changeset-notifications-sms>{{changeset-get this.changeset "notifications.sms"}}</span></h2>
 <br />
 <h2>Model Email: <span data-test-model-user-name>{{this.model.user.name}}</span></h2>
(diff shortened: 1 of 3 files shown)


tests/dummy/app/templates/components/changeset-form.hbs
```diff
@@ -1,10 +1,14 @@
...
 <h2>Changeset Email: <span data-test-changeset-user-email>{{this.changeset.user.email}}</span></h2>
 <h2>Changeset cid: <span data-test-changeset-cid>{{this.changeset.cid}}</span></h2>
+<h2>Changeset notifications email: <span data-test-changeset-notifications-email>{{this.changeset.notifications.email}}</span></h2>
+<h2>Changeset notifications sms: <span data-test-changeset-notifications-sms>{{this.changeset.notifications.sms}}</span></h2>
 <br />
 <h2>Model Email: <span data-test-model-user-name>{{this.model.user.name}}</span></h2>
 <h2>Model Email: <span data-test-model-user-email>{{this.model.user.email}}</span></h2>
 <h2>Model cid: <span data-test-model-cid>{{this.model.cid}}</span></h2>
+<h2>Model notifications email: <span data-test-model-notifications-email>{{this.model.notifications.email}}</span></h2>
+<h2>Model notifications sms: <span data-test-model-notifications-sms>{{this.model.notifications.sms}}</span></h2>
```
snewcomer: Can we confirm that using `changeset-get` helper in `master` makes the tests pass?

josemarluedke: Confirmed. Rebased with master and changed to use `changeset-get`. Bug still there.

snewcomer:
@josemarluedke I think I know what we can do.  We check if the old content value is equal to what we want to set and if so, don't do anything.  I'll see if we can refactor this.

snewcomer:
#416 partially fixed your issues.  I think there might be something else we need to do!

snewcomer:
Ok everything should be g2g on master!  Can't thank you enough for your contributions.  These were also issues in 2.x and 1.x so these are very nice additions!

josemarluedke:
Rebased with master, tests did pass locally. Thank you @snewcomer for working on fixing this.

Should we get this tests merged anyway?

Issue #390, opened 2019-12-25
Add ability to call factory function to generate Changeset instead of a class


snewcomer:
`new Changeset(model, validations)` -> `Changeset(model, validations)`

bartocc:
I believe this can be closed now that #418 is merged

PR #418, opened 2020-01-20
Export Changeset class to allow overriding

### LINKED ISSUE #174: Support Warnings
Context
--------
Using `ember-changeset-validations`. Our UI designs have a concept of showing "warning" messages that alert the user that their input may be less than desirable but isn't invalid. When a form field has a warning the data should still be allowed to be set on the model and the user should be allowed to submit successfully. 

Desired Solution
-----------------
Ideally I would like to be able to flag certain validations in my validation maps as "warnings" and when `ember-changeset` sees the warning validation it adds it to the error array, decorates it as a warning so that it can be styled differently, and doesn't make the changeset invalid. This would allow me to do something this in a `validated-field` component
```javascript
// validation-map.js
export default {
  validateNumber({
    isWarning: true,
    integer: true,
  });
};
```

```hbs
{{!-- validated-field.hbs --}}
{{#each errors as |error|}}
  <li class=(if error.warning "is-warning" "is-error")>
    {{error.message}}
  </li>
{{/each}}
```

Current workaround
--------------------
My current workaround for this is to create two changesets one with a "warnings" validation map and one with an "errors" validation map. Then in my form component I merge the errors from the two changesets and decorate them as "errors" or "warnings".

snewcomer:
This PR would allow users to import the Changeset and extend from it.

```js
import { EmberChangeset } from 'ember-changeset
```

And use like so

```js
changeset(model, validationFn, validationMap, { changeset: ModifiedChangesetKlass })
```

Also the recommended way to generate new changesets is through a factory function instead of a class and the README has been updated.  I'm not sure why the original implementation had a class w/ constructor but the factory function seems better. No breaking changes though.  

```js
import { Changeset } from 'ember-changeset
```

close #174 #390

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ See also the [plugins](#plugins) section for addons that extend `ember-changeset
 #### tl;dr
 
 ```js
-let changeset = new Changeset(user, validatorFn);
+let changeset = Changeset(user, validatorFn);
 user.get('firstName'); // "Michael"
 user.get('lastName'); // "Bolton"
 
@@ -53,7 +53,7 @@ user.get('lastName'); // "Bob"
 
 ## Usage
 
-First, create a new `Changeset` using the `changeset` helper or through JavaScript:
+First, create a new `Changeset` using the `changeset` helper or through JavaScript via a factory function:
 
 ```hbs
 {{! application/template.hbs}}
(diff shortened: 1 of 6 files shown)


snewcomer:
@josemarluedke fyi

PR #421, opened 2020-01-22
add fallback for getOwnPropertyDescriptors

### LINKED ISSUE #420: IE11 does not have `getOwnPropertyDescriptors`
##  Version
above 2.1.2

## Steps to reproduce
use ember-chageset in IE11 

## Expected Behavior
should work probably

## Actual Behavior
throws `Object.getOwnPropertyDescriptors` not defined error 

probably should this kind of fallback https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/utils/lib/get-own-property-descriptors.ts

bekzod:
Closes #420

### DIFF
--- a/addon/utils/assign.ts
+++ b/addon/utils/assign.ts
@@ -1,6 +1,23 @@
+let getOwnPropertyDescriptors: (obj: { [x: string]: any }) => { [x: string]: PropertyDescriptor };
+
+if (Object.getOwnPropertyDescriptors !== undefined) {
+  getOwnPropertyDescriptors = Object.getOwnPropertyDescriptors;
+} else {
+  getOwnPropertyDescriptors = function(obj: object) {
+    let desc: Record<string, any> = {};
+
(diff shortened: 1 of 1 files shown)


addon/utils/assign.ts
```diff
@@ -1,6 +1,23 @@
+let getOwnPropertyDescriptors: (obj: { [x: string]: any }) => { [x: string]: PropertyDescriptor };
+
+if (Object.getOwnPropertyDescriptors !== undefined) {
+  getOwnPropertyDescriptors = Object.getOwnPropertyDescriptors;
+} else {
+  getOwnPropertyDescriptors = function(obj: object) {
```
snewcomer: `obj: Record<string, any>`

I think we need this so accessing `descriptors[key]` doesn't fail with an index signature error.

bekzod: thanks, fixed

Issue #424, opened 2020-02-06
Missing types in v3


josemarluedke:
Prior versions exported a few types that are useful.

```
ValidationResult,
ValidatorMapFunc,
ValidatorAction,
ChangesetDef
```

It seems that instead of using `ChangesetDef`, we should not use `BufferedChangeset`. Confirm?

Also, not having a type definition for the module is a pain for typescript users.

For example, I'm importing `import Changeset from 'ember-changeset';`  Typescript will complain that there is no type definition for this module. We should include an `index.d.ts`.

## Version
3.0.1

snewcomer:
@josemarluedke 👋 I'm currently trying to figure out the best way to do this.  Would appreciate any help or thoughts!

So the types are in `validated-changeset/types` or `validated-changeset/types/types`.  Should we re-export from `ember-changeset/types`?  Does this work for you?  What should be included in the top level `index.d.ts`?

snewcomer:
@josemarluedke Lmk if the latest 3.0.3 works for you!

josemarluedke:
I'm testing out the last version, I will keep posting as I find things. 

Here is the first one: 

Interestingly, I had to import the types from `ember-changeset/addon/types` and not `ember-changeset/types`.

josemarluedke:
Type for `ValidatorMap` is missing. This is useful when working with the `lookupValidator` from `ember-changeset-validations`.

snewcomer:
ok try 3.0.4!

josemarluedke:
@snewcomer I believe v3.0.4 fixes all my issues at the moment. 

Here is a commit in a project applying some changes to accommodate the new types: https://github.com/josemarluedke/frontile/pull/30/commits/0a1bd2d5ca34b8d2c01a5e047b9058291164bcfb

I will close this issue and if I find anything else I will let you know. Thank you for working on all these things.

PR #427, opened 2020-02-07
enhancement declare typings

snewcomer:
After extracting `validated-changeset` to it's own TS library, we still need to allow consumers to reference the shape of their data and possibly extend and create their own `Changeset` class.

ref #424

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ export default class Foo extends Component {
 
   constructor() {
     super(...arguments);
+
     this.changeset = new Changeset(obj);
   }
 }
--- a/README.md
+++ b/README.md
@@ -156,16 +156,23 @@ Be sure to call `validate()` on the `changeset` before saving or committing chan
 ## Types
 
 ```ts
+import Component from '@glimmer/component';
 import { BufferedChangeset } from 'ember-changeset/types';
 import { Changeset } from 'ember-changeset';
 
-export default class Foo extends Component {
-  changeset?: BufferedChangeset;
+interface Args {
+  user: {
+    name: string;
+    age: number;
+  }
+}
 
-  constructor() {
-    super(...arguments);
+export default class Foo extends Component<Args> {
+  changeset: BufferedChangeset;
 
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import {
+  ValidationResult,
+  ValidatorMapFunc,
+  ValidatorAction
+} from 'validated-changeset/dist/types';
+
+export {
+  BufferedChangeset,
+  Changeset,
```
josemarluedke: I believe this should also be exported as default. So that `import Changeset from 'ember-changeset` works.

Not sure if this is the right thing to export, but it should have a default export.

snewcomer: Good idea.  One thing I am trying to figure out is if we need a `declare module` statement for it to actually work...

snewcomer: @josemarluedke What do you think of the PR in it's current state?

README.md
```diff
@@ -155,14 +155,18 @@ Be sure to call `validate()` on the `changeset` before saving or committing chan
...
+import { Changeset } from 'ember-changeset';
 
 export default class Foo extends Component {
   changeset?: BufferedChangeset;
+
+  constructor() {
+    super(...arguments);
+
+    this.changeset = new Changeset(obj);
+  }
```
chriskrycho: This example needs a couple tweaks!

1. The `...arguments` approach, while the usual for plain JS, will actually not currently type-check.
2. The `changeset` should not be optional, as it's set up in the constructor.
3. Where does `obj` come from? I'd assume it's an argument?

```ts
import Component from '@glimmer/component';
import { BufferedChangeset } from 'ember-changeset/types';
import { Changeset } from 'ember-changeset';

interface Args {
  user: {
    name: string;
    age: number;
  }
}

export default class Foo extends Component<Args> {
  changeset: BufferedChangeset;

  constructor(owner, args) {
    super(owner, args);
    this.changeset = new Changeset(args.user);
  }
}
```

Something like this would be a better example, which would also fully type-check.

PR #428, opened 2020-02-12
enhancement Add types to app folder as well

snewcomer:
I think this will work so you don't have to reach into `addon/types` to import. ember-cli will look for it in the `app/` namespace I believe.  @josemarluedke Lmk if you have the same thought as well.

ref #424


snewcomer:
Or since this is at build time, you would reference `ember-changeset/app/types`?

josemarluedke:
@snewcomer Yeah, in the app folder would not work either. I believe it needs to be on the root folder. eg. `ember-changeset/types/index.d.ts`

snewcomer:
That makes a lot more sense :)

josemarluedke:
I just tested here and moving to root does work correctly.

types/index.d.ts
```diff
@@ -0,0 +1,20 @@
+import {
```
josemarluedke: Should we also remove this file from `addon/types/index.d.ts`?

Issue #434, opened 2020-03-05
Ember Data `isValid` property is not shadowed by the changesets `isValid`


ohcibi:
I've setup a basic example to demonstrate one problem with changesetting ember data models: https://github.com/ohcibi/changeset-test

The problem is that the models have an `isValid` property that is true by default. When one `get`s the `isValid` property from the changeset of that model the changeset ultimately fails to return the right thing in https://github.com/poteto/ember-changeset/blob/master/addon/index.js#L190

`this[key]` which is `this['isValid']` returns false so the condition for the if in that line fails (and also the `else if` which does the test for the same key in this case) and the changeset then tries to return the `isValid` of the proxied object. Since that one has an `isValid` which is `true` the changeset returns `true` as well.

My suggestion would be to test explicitly for `undefined` instead of just a falsy value. I can PR if thats an appropriate solution.

snewcomer:
@ohcibi Phenomenal!  Yes this is a good catch.  A PR would be welcome!

arenoir:
This is a pretty big bug and think it should be noted in the changelog that this version doesn't work with ember-data.   I have created a quick fix.

PR #439, opened 2020-03-07
prevent changelist properties from being proxyed to underlying object. 

arenoir:
Quick fix for Issue #434

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -187,7 +187,7 @@ export class EmberChangeset extends BufferedChangeset {
 
 
     // return getters/setters/methods on BufferedProxy instance
-    if (this[key]) {
+    if (typeof this[key] !== 'undefined') {
       return this[key];
     } else if (this[baseKey]) {
       const v = this[baseKey];
--- a/tests/unit/changeset-test.js
+++ b/tests/unit/changeset-test.js
@@ -1504,6 +1504,16 @@ module('Unit | Utility | changeset', function(hooks) {
     ]);
   });
 
+
+  test('#changeset properties do not proxy', async function(assert) {
+    dummyModel.setProperties({ name: undefined, password: false, async: true, isValid: 'not proxied'});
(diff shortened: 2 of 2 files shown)


addon/index.js
```diff
@@ -146,6 +158,10 @@ export class EmberChangeset extends BufferedChangeset {
     // 'person.username'
     let [baseKey, ...remaining] = key.split('.');
 
+    if (NOPROXYPROPS.includes(key)) {
+      return this[key];
+    }
+
```
snewcomer: What do you think about this below on line 190?

```js
if (typeof this[key] !== 'undefined') {
```

arenoir: There is a lot of complication here and off the top of my head wouldn't the type of the property `isValid` return [boolean](https://github.com/validated-changeset/validated-changeset/blob/master/src/index.ts#L209)?

snewcomer: Yep and that is the goal.  The simple fix here is to change below from.

```js
if (this[key]) {
```

```js
if (typeof this[key] !== 'undefined') {
```

`isValid` and other properties that also exist on an Ember Data model won't be reflected via accessing through a changeset.    If the user asks the question `isValid`, the answer they should get back is if the changeset is valid.

arenoir:
I’ll update the pull requests and test it on Monday.

snewcomer:
@arenoir For context of a fix I added to `validated-changeset`

https://github.com/validated-changeset/validated-changeset/pull/18

addon/index.js
arenoir: I updated my pull request and rebased.

ohcibi:
@snewcomer isnt the issue in that line https://github.com/poteto/ember-changeset/pull/439/commits/234f66ef23e6e327507ab574e8840ce09f5de0ad#diff-7562eb8eba3c3d86a671045f8c70107eL192 the same? If the test for `key` fails then the test for `baseKey` would fail in the same way as before?

PR #441, opened 2020-03-10
Failing test: `changeset.error.${property}.validation` does not notify changes to observers

jelhan:
Ported the failing test from https://github.com/poteto/ember-changeset-validations/pull/229 as the bug seems to be present in this upstream library as well. For context please also see https://github.com/poteto/ember-changeset-validations/issues/226.


snewcomer:
Mixing a push vs pull based system seems to be the core issue.  I'm going to assume bootstrap-validations will need the change!  Will try and look this week if somebody else doesn't get to it!

PR #449, opened 2020-04-01
Add failing test for input updates in octane

jkeen:
While updating some components for octane/glimmer I found the bug illustrated in this test. Updates from changesets do not get reflected in bound inputs. It seems like internal properties on the changeset aren't being marked as `@tracked` internally? 

I'm not sure yet, but here's the failing test at least to start.


snewcomer:
@jkeen 👋 This is definitely an interesting question.  Luckily, e-c is doing the right thing.  If you add `await settled()` or take advantage of the `waitUntil` helper from @ember/test-helpers, the test will pass.  So @tracked may be async in nature or happen in the next tick.  There also might be some intricacies in the test environment that I am not aware of.  In any case, I'll ask the question tomorrow and follow up!

snewcomer:
@jkeen 

Screen Shot 2020-04-12 at 4 22 40 PM

PR #466, opened 2020-05-03
Leaf detection when merging sibling keys

### LINKED ISSUE #229: 'getting' a Moment instance on a changeset returns a relay instead of that instance

snewcomer:
close #229

👋 @bterkuile

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,8 +1,9 @@
 import { assert } from '@ember/debug';
-import { BufferedChangeset, normalizeObject, pureAssign } from 'validated-changeset';
+import { BufferedChangeset, normalizeObject } from 'validated-changeset';
 import mergeDeep from './utils/merge-deep';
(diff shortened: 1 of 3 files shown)


bterkuile:
Nice! I checked out the `sn/leaf-detectin` branch an replaced the `node_modules/ember-changeset` folder for this code. For the real life test I am performing it works!

PR #469, opened 2020-05-09
Ensure setter is safeSet for tracking context

### LINKED ISSUE #468: Issue with @tracked
I am running into issues setting and consuming a changeset at the same time.

I have the following in a Glimmer component:
```js
// component.ts
export default class Dummy extends Component {
  @tracked changeset: BufferedChangeset;

  get otherProperty() {
    return this.changeset.property
  }
 
  @action
  setProperty() {
    this.changeset.set('property', {});
  }
}
```
```hbs
// template.hbs
{{log this.changeset.property}}
```

I get the following error:
```js
Error: Assertion Failed: You attempted to update [object Object].property to "[object Object]", but it is being tracked by a tracking context, such as a template, computed property, or observer. In order to make sure the context updates properly, you must invalidate the property when updating it. You can mark the property as `@tracked`, or use `@ember/object#set` to do this.
```

As you can see `changeset` is marked as `@tracked`, but that didn't seem to help.

If I just set the value, without consuming it in the template or another property, I don't get the error.

snewcomer:
close #468

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -20,10 +20,12 @@ export class EmberChangeset extends BufferedChangeset {
 
   // override base class
   getDeep = safeGet;
-
-  // override base class
+  setDeep = safeSet;
   safeGet(obj, key) {
     return safeGet(obj, key);
+  }
+  safeSet(obj, key, value) {
+    return safeSet(obj, key, value);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


snewcomer:
Not the right approach yet.  Investigating `validated-changeset` and using `safeSet` overridable property instead of plain JS setter

PR #471, opened 2020-05-12
Safe set for setDeep

### LINKED ISSUE #468: Issue with @tracked

snewcomer:
close #468

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -26,9 +26,6 @@ export class EmberChangeset extends BufferedChangeset {
   safeGet(obj, key) {
     return safeGet(obj, key);
   }
-  safeSet(obj, key, value) {
-    return safeSet(obj, key, value);
-  }
 
   /**
    * Manually add an error to the changeset. If there is an existing
--- a/addon/index.js
+++ b/addon/index.js
@@ -28,6 +28,9 @@ export class EmberChangeset extends BufferedChangeset {
   safeGet(obj, key) {
     return safeGet(obj, key);
   }
+  safeSet(obj, key) {
+    return safeSet(obj, key);
+  }
 
   /**
    * Manually add an error to the changeset. If there is an existing
--- a/addon/index.js
+++ b/addon/index.js
@@ -28,8 +28,8 @@ export class EmberChangeset extends BufferedChangeset {
   safeGet(obj, key) {
     return safeGet(obj, key);
   }
-  safeSet(obj, key) {
-    return safeSet(obj, key);
+  safeSet(obj, key, value) {
+    return safeSet(obj, key, value);
   }
 
   /**
--- a/addon/index.js
+++ b/addon/index.js
@@ -19,11 +19,15 @@ export class EmberChangeset extends BufferedChangeset {
   @tracked '_content';
 
   // override base class
+  // DO NOT override setDeep. Ember.set does not work with Ember.set({}, 'user.name', 'foo');
   getDeep = safeGet;
 
   // override base class
   safeGet(obj, key) {
     return safeGet(obj, key);
+  }
+  safeSet(obj, key, value) {
+    return safeSet(obj, key, value);
   }
 
   /**
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "ember-auto-import": "^1.5.2",
     "@glimmer/tracking": "^1.0.0",
     "ember-cli-babel": "^7.19.0",
-    "validated-changeset": "~0.4.4"
+    "validated-changeset": "0.4.1"
   },
   "devDependencies": {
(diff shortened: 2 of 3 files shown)


basz:
Not sure if this is *the* way to test this PR... Just giving some feedback...

// package.json
```
    "ember-bootstrap-changeset-validations": "^3.0.0",
    "ember-changeset": "git+https://github.com/poteto/ember-changeset.git#sn/safe-set-deep",
    "ember-changeset-validations": "^3.3.1",
```

Unfortunatly still seeing issue #468

snewcomer:
@basz Would you mind trying again?  If you don't mind inspecting the node_modules and see if these changes made it in there?

basz:
Hmm. Looks like the values are set correctly now... 🎉 However, for nested keys my validations don't seem to run... That could be because I'm coming from the v2 releases of

```
"ember-bootstrap-changeset-validations": "^2.0.0"
"ember-changeset": "^2.1.2",
"ember-changeset-validations": "^2.1.0",
```

anything obvious I should be aware of?

snewcomer:
Awesome!  What does it mean that nested validations don't run?  If `user.email` are you setting `email` directly or setting `user` and expecting `email` validations to run?

basz:
I'm confused.

should the validationsMap for nested keys be like version 2.

```js
{
  key1: [validations],
  key2.nested1: [validations],
}
```

or (nested pojo's)

```
{
  key1: [validations],
  key2: {
    nested1: [validations],
  }
}
```

snewcomer:
nested pojo's.  Using the dot separated version was very limiting for cases detailed in places in #379

basz:
ok thanks. 

We should change that on the README... Search for "specify nested keys with dot delimiters"...

basz:
That solved the validation when I edit a field. When I submit the form I do the following

```
this.changeset.validate().then(() => {
      alert(this.changeset.get('isInvalid'));

      console.log(this.changeset.get('errors'));
    });
```
Errors is then empty, alert is false, whilst it shouldn't...

snewcomer:
@basz If there was a minimal repo of this issue, would be happy to take a look!

basz:
There seems to be a problem (or at least a difference between v2 and v3) with detecting the keys to be validated.

When calling `this.changeset.validate()` none of the validations aren run (which used to be all validtions). when I do `this.changeset.validate('consumer.name.givenname')` just that one and isInvalid becomes true...

basz:
I could do something like https://github.com/hughsk/flat on the validationMap and feed it into the validate method... But I'll wait until you confirm it isn't something validate should do by itself and is indeed a difference between v2 and v3...

snewcomer:
@basz Interesting!  We do have tests for `validate/0`.  I'd be curious on what part of `changeset.validate()` isn't working.

basz:
Yeah, I went through the tests an found it...

It's not working on nested keys. So given;

```js
{
  'key1': [validations],
  'key2': {
    'nested1': [validations],
  }
}
```
and then calling `this.changeset.validate()` only runs the 'key1' validations... Not sure if that is intended though... (the test '#validate/nested validates nested fields immediately' specifically calls the nested key)

basz:
currently working around with

```
validationKeys = Object.keys(flatten(dossierValidations))
      .map(key => key.slice(0, -2))
      .uniq();

this.changeset.validate(...validationKeys);
```

snewcomer:
Oh interesting!  Mind filing an issue?  Nested keys have always been a sore spot.  By moving away from dot separated (v3), we solved most of the issues.  This just seems to be a missed edge by me.

basz:
see https://github.com/poteto/ember-changeset/issues/478

basz:
Thank you @snewcomer for fixing the latest issues and taking care of this package. My upgrade path has been smooth sailing since and ember-changeset is a pleasure to work with...

snewcomer:
Thank you for the `flatten` idea!  Made the solution super simple.

PR #476, opened 2020-05-16
Update README.md

basz:
dot notation to pojo on validationMap

Closes # .


basz:
you guest right! thanks!

PR #480, opened 2020-05-17
Changeset not dirty from hasMany change

allthesignals:
Hi there! This PR adds a test showing changeset isn’t dirty after pushing to a hasmany — not quite intended to actually be merged.

What I want to verify with this PR is:

 - is this intended behavior? 
   - if not, is there a path towards fixing it? anything I could look at?
   - if yes, what're some alternatives?

Happy to help as much as I can :)

### DIFF
--- a/tests/integration/main-test.js
+++ b/tests/integration/main-test.js
@@ -155,8 +155,13 @@ module('Integration | main', function(hooks) {
     let user = this.createUser(userType, true);
     let changeset = Changeset(user);
     let newDog = this.store.createRecord('dog', { breed: 'Münsterländer' });
+
+    assert.equal(changeset.isDirty, false);
+
     let dogs = changeset.get('dogs');
     dogs.pushObjects([newDog]);
+
+    assert.equal(changeset.isDirty, true);
 
     dogs = changeset.get('dogs').toArray();
     assert.equal(dogs[0].get('breed'), 'rough collie');


snewcomer:
Since a changeset's primary goal is as a read-write buffer, since `isDirty` only changes implicitly in the model itself, I would imagine you would want to access it via `model.isDirty`?

allthesignals:
@snewcomer Hi! Thanks for the response. I think that makes sense, but until the changeset is executed and applied to the model, an _ember-data_ model's dirtiness is not known. It's also [documented](https://github.com/poteto/ember-changeset#isdirty) as a distinct thing on the changeset API itself. Other changes (+ `execute()`s) put the changeset in a dirty state, afaict. What do you think?

snewcomer:
Oh this is v-c state!  Yes ok this might be a bug.  Thank you for the test repro!

snewcomer:
@allthesignals So when you call push objects, e-c isn't aware of changes in an external data structure.  You have to replace the whole object.  See #481

allthesignals:
Interesting, thank you @snewcomer!

Just out of curiosity, is https://github.com/poteto/ember-changeset/pull/481 to assert that this is current behavior? So if someone wants to address it they will see those tests fail and update them accordingly?

snewcomer:
Yep!  Basically if you don't go through our Proxy#set, then we can't track changes.  On every update, if you `changeset.set('user', NewUserProxy)`, then we can trap it, mark changes and mark as dirty. 👍 
Lmk if you have other questions!

PR #488, opened 2020-05-29
Add failing test case for belongsTo validation

andreyfel:
If belongsTo is empty and no changes have been made to the changeset
the presence validation must yell at us with "Can't be blank"
validation error when we enforce the validation.

This is a regression which appeared in ember-changeset@3.4.0.
It is not related to ember-changeset-validations.

Presence validation is basically a check !Ember.isEmpty(value).
This test simulates it.

### DIFF
--- a/tests/integration/main-test.js
+++ b/tests/integration/main-test.js
@@ -2,6 +2,10 @@ import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import { set } from '@ember/object';
 import { Changeset } from 'ember-changeset';
+import { isEmpty } from '@ember/utils';
+import ArrayProxy from '@ember/array/proxy';
+import ObjectProxy from '@ember/object/proxy';
+import { get } from '@ember/object';
 
 module('Integration | main', function(hooks) {
   setupTest(hooks);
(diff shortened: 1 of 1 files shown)


snewcomer:
The [ref](https://github.com/validated-changeset/validated-changeset/pull/60/files) PR, we ensure that we don't return CHANGES now.

However as of `0.6.2`, we may still have a problem b/c in your validator, `newValue` will be the Proxy node (b/c it detects an object from CONTENT and thinks you might need to keep accessing deeper levels of the tree).  

I'm wondering if we need to access the underlying CHANGES or CONTENT on the Proxy when sending the user `newValue`?

tests/integration/main-test.js
```diff
@@ -273,4 +277,38 @@ module('Integration | main', function(hooks) {
...
+  function isProxy(o) {
+    return !!(o && (o instanceof ObjectProxy || o instanceof ArrayProxy));
+  }
+
+  async function testBelongsToPresenceValidation(assert, userType) {
+    let user =  this.store.createRecord(userType);
+    function userValidator({ key, newValue }) {
+      const value = unwrapProxy(newValue);
+      if (key === 'profile') {
+        return isEmpty(value) ? 'Cannot be blank' : true;
```
snewcomer, line 294: I believe `newValue` will either be CHANGES or CONTENT.  So if `profile` existed on the underlying model, then `newValue` should technically be "something" b/c it exists on the model.

Are you expecting `newValue` to be undefined?

andreyfel: In case of sync belongsTo() it is null.
In case of async belongsTo() it should be null wrapped in ObjectProxy. There is an unwrapProxy() above, so at this point, it is already unwrapped, so it is null.

andreyfel:
I will test with the latest version to see if it is still the issue.

andreyfel:
It appears that it still fails with validated-changeset@0.6.2
So, maybe we should unwrap the proxy when traversing the tree. Maybe `isObject` util should be smarter and if an argument is a proxy then unwrap it first and assess if it is an object?

tests/integration/main-test.js
snewcomer: Right good point.  At this point though, it is the ObjectTreeNode proxy, not your proxy.  Let me know what you think about the below PR!  Basically we need to unwrap before sending through to the validator function.

https://github.com/validated-changeset/validated-changeset/pull/61

snewcomer:
Just saw this reply.  Lmk where else we can make some improvements by unwrapping the proxy!

validated-changeset/validated-changeset#61

andreyfel:
Hi @snewcomer!
I really appreciate your swift feedback!
I will add a comment to validated-changeset/validated-changeset#61

andreyfel:
This is still an issue even after merging validated-changeset/validated-changeset#61

snewcomer:
Given the inability introspect Proxys we don't have control over, we need an Ember specific way of dealing with ObjectProxy and ArrayProxy

PR #495, opened 2020-06-03
Add failing test case for autotracking of derived state

andreyfel:


tests/integration/misc-test.js
```diff
@@ -0,0 +1,62 @@
...
+import { render, find, settled } from '@ember/test-helpers';
+
+module('Integration | misc', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it re-renders derived state automatically when root state changes', async function(assert) {
+    class MyModel {
+      isOptionOne = false;
+      isOptionTwo = false;
+      isOptionThree = false;
```
snewcomer, line 15: What if you add `@tracked` to these properties?

andreyfel: Sorry, missed that.
It still can be reproduced with `@tracked`

snewcomer:
@andreyfel This looks like a tough one.  When you explicitly set a value (`isOptionTwo`), we look to update the `ERRORS`.  If the error looks like this - `{isOptionSelected: Err}` - then we can't delete it.  Moreover, thinking about `tracked-built-ins`, I'm not seeing off hand how that would help either.  We either push / remove an error explicitly in response to `set`.

Do you have any thoughts as to a solution here?

andreyfel:
You are right that just switching to `tracked-built-ins` won't help here.

One thought I have is to make ERRORS a getter vs a static property and calculate it based on CHANGES.

snewcomer:
I'm only guessing after #509, this is still failing?

andreyfel:
Yep, still failing

PR #496, opened 2020-06-03
Update ember

andreyfel:


andreyfel:
@snewcomer I'm not sure about typescript test file `changeset-get-test.ts`. Is it even executed?

tests/unit/changeset-test.js
```diff
@@ -1688,6 +1688,7 @@ module('Unit | Utility | changeset', function(hooks) {
 
     Object.defineProperty(myChangeset, 'isOptionSelected', {
       get() {
+        // eslint-disable-next-line ember/no-get
         return this.get('isOptionOne') || this.get('isOptionTwo') || this.get('isOptionThree');
```
snewcomer, line 1692: Can we ✂️ `this.get` and do `this.isOptionOne...`?

andreyfel: No, it doesn't work without this.get

snewcomer: Sorry can you explain why it doesn't work with `this[property]`?

andreyfel: I think this is because at this point `this` is not a `Proxy`, it is a `BufferedChangeset`. So, if I don't explicitly write `this.get` it doesn't call BufferedChangeset.get.

PR #509, opened 2020-06-15
Add dependentKeyCompat to isValid/isInvalid

### LINKED ISSUE #455: isValid / isInvalid are not tracked
It is natural to want to use `isValid` or `isInvalid` in a template:

`<button disabled={{changeset.isInvalid}}>SAVE</button>`

These should therefore be tracked so that they will update in a template.

snewcomer:
close #455

https://api.emberjs.com/ember/3.15/functions/@ember%2Fobject%2Fcompat/dependentKeyCompat

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,6 +1,6 @@
 import { assert } from '@ember/debug';
 import { dependentKeyCompat } from '@ember/object/compat';
-import { BufferedChangeset } from 'validated-changeset';
+import { BufferedChangeset, getKeyValues } from 'validated-changeset';
 import ArrayProxy from '@ember/array/proxy';
 import ObjectProxy from '@ember/object/proxy';
 import mergeDeep from './utils/merge-deep';
@@ -48,7 +48,7 @@ export class EmberChangeset extends BufferedChangeset {
    */
   @dependentKeyCompat
   get isValid() {
-    return super.isValid(...arguments);
+    return super.isValid;
   }
 
   /**
@@ -57,7 +57,16 @@ export class EmberChangeset extends BufferedChangeset {
    */
   @dependentKeyCompat
   get isInvalid() {
-    return super.isInvalid(...arguments);
+    return !this.isValid;
+  }
+
+  /**
+   * @property isDirty
+   * @type {Boolean}
+   */
+  @dependentKeyCompat
+  get isDirty() {
+    return !this.isPristine;
   }
 
   /**
--- a/addon/index.js
+++ b/addon/index.js
@@ -57,7 +57,16 @@ export class EmberChangeset extends BufferedChangeset {
    */
   @dependentKeyCompat
   get isInvalid() {
-    return !this.isValid;
+    return super.isInvalid;
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


BryanCrotaz:
Aha! Are there any other places that this needs to be added?

snewcomer:
That is a good question.  Added `isDirty` and `isPristine` as well!

snewcomer:
Trying to get a release out.  npm is giving me a 404...

alexlafroscia:
Super cool! Thanks for digging into it @snewcomer. 

I already refactored my code into native getters, but I’m try out reverting that tomorrow and see if it works again!

snewcomer:
`v3.6.1`!!

alexlafroscia:
Doesn't seem like this worked for me 😦 I updated the PR I'm working on to `3.6.1` and reverted my changes that converted the use of the CP macros to getters, and now my tests fail again.

snewcomer:
Sad :(.  There are a lot of moving parts.  We probably reached the boundary of what we can do in the `3.x` series. For users that want to maintain this syntax, they probably need to stick to the `2.x` series.  

These macro decorators may help for certain situations to get one off of these computed property macros.

https://github.com/pzuraq/macro-decorators

alexlafroscia:
> For users that want to maintain this syntax, they probably need to stick to the 2.x series.

I think this as a pretty big problem... I'm trying to get my app upgraded to Ember 3.13+ and ran into issues with `ember-changeset@v2` there. Upgrading to V3 as part of the app moving to Ember 3.13+ seems to be necessary. Requiring that every computed property that depends on some state from a changeset _must_ be re-written into a getter makes this a really massive task, as we have have this kind of stuff all over the place

BryanCrotaz:
Seems like we need to get an Ember core team person involved here. According to the docs this should work

alexlafroscia:
Just for my own sanity, I added a failing test to `ember-changeset` to make sure that things indeed don't work.

Talking briefly to @pzuraq on Discord, it seems like adding `@tracked` to an existing property on a base class does not work, which is what `ember-changeset` seems to be doing

https://github.com/poteto/ember-changeset/blob/c570bda5e95b0374a79fe9b2f0e75cc1bcc76061/addon/index.js#L25-L27

Where the base class also defines these:

https://github.com/validated-changeset/validated-changeset/blob/40676661e52ec2733ff1606607d9f68d5781a7c6/src/index.ts#L114-L116

Chris's recommendation was to create a wrapper class that handles the "tracked" behavior, which I'm playing with locally right now.

snewcomer:
If that is what we need, then happy to accept a PR!!

alexlafroscia:
Cool -- I'm going to see if I can break apart my Ember 3.13+ upgrade from the `ember-changeset` one to make the whole thing easier to land. If I can get my test passing, I'll definitely make a PR to upstream the fix! So far no luck, but I'm going to keep trying.

alexlafroscia:
I didn't end up getting into it -- our codebase is still using the older release of `ember-changeset` for now

PR #510, opened 2020-06-15
rm notifyPropertyChange from specific spots

snewcomer:
Testing this out.  With `@tracked`, I don't know if we need these or not.


addon/index.js
```diff
@@ -87,10 +86,6 @@ export class EmberChangeset extends BufferedChangeset {
   addError(key, error) {
     super.addError(key, error);
```
snewcomer, line 87: https://github.com/validated-changeset/validated-changeset/blob/40676661e52ec2733ff1606607d9f68d5781a7c6/src/index.ts#L559

For example, `addError` will set the `ERRORS` object and since it is tracked, anything consuming it will pull in a new value

Issue #512, opened 2020-06-18
pendingData property to get resulting object with changes applied


RuslanZavacky:
Have a feature request?
=======================

I am working on a use case, where ember-changeset is not the single solution to validate data. We use JSON Schema to validate fields, and some of those fields are dynamic. So their validation is derived from the server, when we've got that schema. 

Our validation process goes through 2 stages: 
(1) validate changeset 
(2) validate through JSON schema

To pass proper data to JSON schema for validation, we have to call `changeset.execute().data` as we need a current set of fields.

But here is an issue: if customer pressed Submit on form once, and there are errors in Changeset, it won't give me latest execute() to apply, those making JSON schema validation incorrect, as it won't run on latest data that customer entered.

```
/**
   * Executes the changeset if in a valid state.
   *
   * @method execute
   */
  execute() {
    if (this.isValid && this.isDirty) {
      let content = this[CONTENT];
      let changes = this[CHANGES];
      // we want mutation on original object
      // @tracked
      this[CONTENT] = mergeDeep(content, changes, { safeGet, safeSet });
    }

    return this;
  }
```

--

Proposal would be, to add a `force` mode to execute, if it is passed - it would apply the data, regardless if it is valid or not. Something like:

```
  execute({ force = false } = {}) {
    let shouldExecute = (this.isValid || force) && this.isDirty;

    if (shouldExecute) {
      let content = this[CONTENT];
      let changes = this[CHANGES];
      // we want mutation on original object
      // @tracked
      this[CONTENT] = mergeDeep(content, changes, { safeGet, safeSet });
    }

    return this;
  }
```

Another option could be to introduce a separate method that can return full object with pending changes applied.

Is there something I might be missing in such approach? If not, I can open a PR to add it.

RuslanZavacky:
Update: Approach changed in PR, instead of changing `execute()` method, new `pendingData` property is introduced. For now it would work only for POJOs, and if you want to merge two ember-data models, you still have to use `execute()` method.

snewcomer:
3.7.0!

RuslanZavacky:
Thank you!

PR #513, opened 2020-06-18
#512 Allow to pass force to execute, to apply changes if content is not valid

RuslanZavacky:
#512 - this introduces ability to forcefully apply changes to the object via execute method, even if validation wasn't successful.


addon/index.js
```diff
@@ -152,10 +152,13 @@ export class EmberChangeset extends BufferedChangeset {
   /**
    * Executes the changeset if in a valid state.
    *
+   * @param {{ force: boolean }}
    * @method execute
    */
-  execute() {
-    if (this.isValid && this.isDirty) {
+  execute({ force = false} = {}) {
```
snewcomer, line 158: `r/{ force = false}/{ force = false }`

but b/c no lint rule, not necessary if you don't feel like it!

RuslanZavacky: My bad, will fix :)

RuslanZavacky:
@snewcomer on the passing object vs variable to a function - I do prefer passing objects, as when codebase is JS, it would allow auto-complete with suggestion of a variable, and then in your code base you'd have something like:

```
execute({ force: true });
```

vs

```
execute(false);
```

Also in case of future additions, having object instead of variable allows for easier addition of functionality. As if you need a 2nd param in that method, `execute(false, true)` might not look that great :) vs `execute({ skipCheck: true });` or something like that.

But again, it is up to peoples choice, if project decides to use only variables - it is also a way forward, if that works for the project :)

RuslanZavacky:
@snewcomer my one question is if that change makes more sense, vs creating a new method (or variable) that would always have up to date data object data + changes together?

snewcomer:
> up to date data object data + changes together

I like the idea!  One notorious issue though is deep cloning and apply those changes to the underlying model (esp when it comes to ED models and the like)

RuslanZavacky:
@snewcomer I will try it out, as I guess it will be very similar to execute() method, but instead of applying data to the existing object, it might return combined object (or apply into intermediate one, need to test out what's better). 

Don't merge this PR please, I'll try it out tomorrow :)

snewcomer:
I guess my intention in the previous comment was that trying to deep clone (to not mutate the existing `model`) the object presents a whole host of issues.

RuslanZavacky:
@snewcomer introduced `pendingChanges` getter, I think I like that approach more, as it is less invasive and don't change existing APIs. If you think it is 👍 I'll add documentation and then PR might be good to go :)

tests/unit/changeset-test.js
```diff
@@ -1057,6 +1057,60 @@ module('Unit | Utility | changeset', function(hooks) {
...
+  });
+
+  test('#pendingChanges up to date after saved changeset modified again', async function(assert) {
+    let dummyChangeset = Changeset(dummyModel);
+    dummyChangeset.set('name', 'foo');
+
+    assert.equal(get(dummyModel, 'name'), undefined, 'precondition');
+    assert.equal(get(dummyChangeset, 'isValid'), true, 'can be valid');
+
+    assert.equal(get(dummyModel, 'name'), undefined, 'original data is not applied');
```
snewcomer, line 1096: Can we move this assertion after the `pendingData` check?

RuslanZavacky: 👍

RuslanZavacky:
hm, do I need to add `dependentKeyCompat` to that getter as well?

tests/unit/changeset-test.js
```diff
@@ -1057,6 +1057,60 @@ module('Unit | Utility | changeset', function(hooks) {
...
+      dummyModel.getProperties('name', 'exampleArray'),
+      dummyChangeset.pendingData,
+      'pending changes equal to original data'
+    );
+
+    dummyChangeset.set('name', 'bar');
+
+    assert.equal(get(dummyChangeset.pendingData, 'name'), 'bar', 'pending data is returned in pendingChanges');
+    assert.equal(get(dummyModel, 'name'), 'foo', 'original data is not modified');
+  });
```
snewcomer, line 1112: Great job so far!  We probably need a test with an EmberData model - both pendingData on a belongsTo property and pendingData by swapping out a belongsTo relationship for another.  I'm skeptical the merge-deep util I made won't mutate the underlying model in certain cases.  Perhaps it doesn't; however, I don't have enough unit tests around `merge-deep` to be 100% confident.

RuslanZavacky: Good point, will add some more test cases

snewcomer: (I like this change btw.  Great idea in place of the `force` option.

RuslanZavacky:
Working with ember-data models unfortunately doesn't work the same way...

My initial test was with simple objects (the ones that we use in our use-case). When you want a deep copy of objects, without modifying of source object - I do merge `{}` <- `<source object>`, then `new object` <- `target object`. That works, and give me proper combined object, without modifying source.

With ember-data, such schema doesn't work. We need to get a clone of the source model first, and then merge stuff into it. As I don't have much experience with ember-data, I don't know what issues can arise from cloning a model (even how to clone it properly?) then merge stuff in, and then discard it, so it accidently is not going to be saved through store.

So far, that's what happens - instead of just value "Zoe", I get Change object with value inside, but that's not what I expect to get when fetching model.

Screenshot 2020-06-20 at 14 52 16


@snewcomer maybe you have some ideas?

snewcomer:
@RuslanZavacky Perhaps we can get up a test? 

Some notes I am suspecting that makes `pendingChanges` hard for e-d objects.  Detecting if an Ember Data object isn't straightforward/even possible?  It is sort of an estimate.  Also I put up this [PR](https://github.com/poteto/ember-changeset/pull/514) to show that we could use something like `try/catch` if we got this far 👇 since a max call stack error will be thrown for certain cases (like the one in the PR).

```
try {
  pendingChanges = mergeDeep(pendingChanges, changes, { safeGet, safeSet });
} catch(e) {
  throw new Error('Unable to determine `pendingChanges` with your data');
}
```

We could document and throw and error to developers.

RuslanZavacky:
@snewcomer I've added 2 test cases that cover ember-data model, with `todo` mark, as they fail :) 

I am not sure if in this case try/catch helps. In my test cases - merging works without errors, but it doesn't give me back models, rather than POJO with some Change objects inside.

snewcomer:
> doesn't give me back models

Ah right.  Might have to do something like this.  (this is what makes this complicated :( but perhaps we can come to a solution eventually)

```
let pendingChanges = mergeDeep(Object.create(Object.getPrototypeOf(content)), content, { safeGet, safeSet });
```

RuslanZavacky:
@snewcomer oo, nice. Changed to that. That's even more correct than I was merging into empty object. But yea, ember-data objects won't work with my implementation anyways. Tried to adapt tests, but ht the circular dependency, that you've added fix for. 👍 for the fix.

I think that as much as we can add it for now? Should I add docs for this getter and explain it limitations?

tests/unit/changeset-test.js
```diff
@@ -1057,6 +1060,128 @@ module('Unit | Utility | changeset', function(hooks) {
...
+
+    dummyChangeset.set('profile.firstName', 'Zoe');
+
+    assert.equal(get(dummyChangeset, 'isPristine'), false, 'changeset is not pristine as there is a change');
+
+    assert.equal(dummyChangeset.get('pendingData.profile.firstName'), 'Zoe', 'Model belongsTo property should be updated')
+    assert.equal(dummyChangeset.get('pendingData.profile.lastName'), 'Ross', 'Existing property should stay the same')
+
+    assert.equal(dummyChangeset.get('data.profile.firstName'), 'Bob', 'Original model property should stay without changes')
+    assert.equal(dummyChangeset.get('data.profile.lastName'), 'Ross', 'Original model property should stay without changes')
```
snewcomer, line 1140: 👍 Would be good to assert `mockUserModel` properties as well

tests/unit/changeset-test.js
```diff
@@ -1057,6 +1060,128 @@ module('Unit | Utility | changeset', function(hooks) {
...
+    assert.equal(get(mockUserModel, 'profile.firstName'), 'Zoe', 'belongsTo property is model has initial value');
+    assert.equal(get(dummyChangeset, 'profile.firstName'), 'Bob', 'belongsTo changeset is updated');
+    assert.equal(get(dummyChangeset.pendingData, 'profile.firstName'), 'Bob', 'belongsTo pendingData has new value');
+
+    await dummyChangeset.save();
+
+    assert.equal(get(mockUserModel, 'profile.firstName'), 'Bob', 'belongsTo property is updated');
+    assert.equal(get(dummyChangeset, 'profile.firstName'), 'Bob', 'belongsTo changeset is updated');
+    assert.equal(get(dummyChangeset.pendingData, 'profile.firstName'), 'Bob', 'belongsTo pendingData is updated');
+  });
```
snewcomer, line 1183: What isn't working now with e-d objects?  If you hit the circular dependency issue, wouldn't we have an `assert.throws` assertion?

RuslanZavacky: circular dependency issue happens in one test, but not in another though. And `assert.throws` didn't work in that case. 2nd tests was throwing exception correctly. I'll test it out more.

tests/unit/changeset-test.js
RuslanZavacky: Sure, 👍

RuslanZavacky: Done

tests/unit/changeset-test.js
RuslanZavacky: Weird, I wasn't able to reproduce it. I think what I was seeing:

I've added console.error() in that new try/catch block, to see the error. Then I've added logging in mergeDeep, to see which keys are failing. And then when I saw that it fails with `null`, I've added `possible !== null` and after that it was just going in loops, without breaking through max call stack. But I think it was my failed experiments :) Now everything seems to be fine.

RuslanZavacky:
@snewcomer addressed all comments and updated [documentation](https://github.com/poteto/ember-changeset/blob/88673200c421483fc37d2ce14cbc06bd329236bf/README.md#pendingData).

If you have some other comments, I will be happy to address them :) 

Thank you for your help 👍

tests/unit/changeset-test.js
```diff
@@ -1057,6 +1060,137 @@ module('Unit | Utility | changeset', function(hooks) {
...
+    assert.equal(get(dummyChangeset, 'profile.firstName'), 'Zoe', 'belongsTo changeset has new value');
+    assert.equal(get(dummyChangeset.pendingData, 'profile.firstName'), 'Zoe', 'pendingData should have a new value');
+
+    let updatedMockProfileModel = store.createRecord('profile');
+
+    dummyChangeset.set('profile', updatedMockProfileModel);
+
+    assert.equal(get(mockUserModel, 'profile.firstName'), 'Zoe', 'belongsTo property is model has initial value');
+    assert.equal(get(dummyChangeset, 'profile.firstName'), 'Bob', 'belongsTo changeset is updated');
+    assert.equal(get(dummyChangeset.pendingData, 'profile.firstName'), 'Bob', 'belongsTo pendingData has new value');
```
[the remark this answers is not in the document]
RuslanZavacky: This test is marked as `todo`. In this case, when there is an error, it will be marked as "passed". When the functionality will be fixed/adapted, and it will be working - this test will fail :) I've created 2 tests specifically, one that has assert.throws another with todo, so they would fail in one or another case.

tests/unit/changeset-test.js
```diff
@@ -1057,6 +1060,137 @@ module('Unit | Utility | changeset', function(hooks) {
...
+    });
+
+    let dummyChangeset = Changeset(mockUserModel);
+    assert.equal(get(dummyChangeset, 'isPristine'), true, 'changeset is pristine');
+
+    dummyChangeset.set('profile.firstName', 'Zoe');
+
+    assert.equal(get(mockUserModel, 'profile.firstName'), 'Bob', 'belongsTo property is not modified');
+    assert.equal(get(dummyChangeset, 'profile.firstName'), 'Zoe', 'belongsTo changeset is modified');
+    assert.equal(get(dummyChangeset.pendingData, 'profile.firstName'), 'Zoe', 'pending data should give updated values');
```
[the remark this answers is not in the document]
RuslanZavacky: Same as https://github.com/poteto/ember-changeset/pull/513#discussion_r444282576

tests/unit/changeset-test.js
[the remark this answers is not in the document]
snewcomer: Perfect!  I missed that.  At least we have a test for the circular issue!

PR #514, opened 2020-06-20
Throw error if merging e-d objects

snewcomer:
ref #513.  Also for a potential solution for `pendingChanges`

This error is throw b/c of a Max Call Stack

### DIFF
--- a/addon/utils/merge-deep.js
+++ b/addon/utils/merge-deep.js
@@ -140,7 +140,7 @@ export default function mergeDeep(target, source, options = {}) {
     try {
       return mergeTargetAndSource(target, source, options);
     } catch(e) {
-      throw new Error('Unable to determine `mergeDeep` with your data');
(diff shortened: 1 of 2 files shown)


addon/utils/merge-deep.js
```diff
@@ -137,6 +137,10 @@ export default function mergeDeep(target, source, options = {}) {
   } else if (sourceIsArray) {
     return source;
   } else {
-    return mergeTargetAndSource(target, source, options);
+    try {
+      return mergeTargetAndSource(target, source, options);
+    } catch(e) {
+      throw new Error('Unable to `mergeDeep` with your data');
+    }
```
RuslanZavacky, line 144: As a code-golf, maybe you can drop `else` clause here? As it is the last code path?

For the error message, do you think - maybe it worth to specify that it is due to the merging of two ember-data models, that result in circular merge sequence?

PR #516, opened 2020-06-25
Failing test: manually-reverted values are too "sticky"

mwpastore:
Consider a template with a form to manipulate a property of a changeset. If you use template actions to change the property away from and subsequently back to its pristine value, it can't be changed again using a template action.

### DIFF
--- a/tests/integration/components/changeset-test.js
+++ b/tests/integration/components/changeset-test.js
@@ -537,5 +537,69 @@ module('Integration | Helper | changeset', function(hooks) {
     assert.equal(find('h1').textContent.trim(), 'J B', 'should update observable value');
     assert.notOk(find('#error-paragraph'), 'should skip validation');
   });
+
+  test('can re-apply a "manually-reverted" change', async function(assert) {
+    /*
+     * Whether you use an inline `{{hash}}' here or a "real" Object or
+     * EmberObject or ES6 instance or whatever--doesn't matter.
+     *
+     * Whether you use `{{on "click" ..}}' or `onclick={{..}}' or a classic
+     * `{{action}}' "modifier"--doesn't matter.
+     *
+     * Whether you use `{{fn (changeset-set ..)}}' here or `{{action (mut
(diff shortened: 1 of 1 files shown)


mwpastore:
I traced it all the way through to where validated-changeset calls Ember.set on the changes object (from [here](https://github.com/validated-changeset/validated-changeset/blob/6809d934f1dc26ca0de917b9574bddc5cc4b4d33/src/index.ts#L785-L787) to [here](https://github.com/validated-changeset/validated-changeset/blob/6809d934f1dc26ca0de917b9574bddc5cc4b4d33/src/utils/set-deep.ts#L55)). The first time around, when you use a template action to set `aProperty` to `bar`, it calls:

```javascript
options.safeSet(
  {}, // target
  "aProperty", // path
  { value: "bar" } // value
)
```

After this operation, `target` (changes) looks like: `aProperty: { value: "bar" }`. 

The second time around, after setting `aProperty` to `foo` and then to `qux`, it calls:

```javascript
options.safeSet(
  {}, // target
  "aProperty", // path
  { value: "qux" } // value
)
```

After this operation, `target` (changes) looks like: `{}`. 

So for whatever reason Ember.set is buffering or otherwise swallowing this operation.

mwpastore:
Welp, there's either too much property notification happening or not enough. 😂 If I modify setDeep to just do `target[path] = value` instead of calling safeSet, this test passes. But it breaks other things (as one might expect). Not sure where to go from here... feeling a bit out over my skis!

snewcomer:
@mwpastore I'm digging into Ember.set and it looks like it is only a problem in DEBUG mode (some WeakMap logistics with Embe.set in dev/test mode).  Will keep digging and paste when/if I get a valid test case!

snewcomer:
Fixed by https://github.com/emberjs/ember.js/pull/18961

Looks like this was fixed 2 months ago here ☝️!!!  This looks like 3.20-beta series.  Checking if it is something can be backported but I'm guessing not.

mwpastore:
Sure enough, my test passes on ember-beta! Thanks @snewcomer!

Issue #499, opened 2020-06-05
property change notification too broad


basz:
Hi,

I'm noticing that changing one property path also triggers updates to other sibling properties.

a setter (or didReceivedAttrs) is triggered for components like so when a different property in the changeset is changed...

```
<MyComponent @value={{changeset-get this.changeset "aProperty"}}/>
```

ember-changeset@^3.5.3
validated-changeset@~0.6.5

snewcomer:
@basz Happen to have more information?

basz:
Checkout: https://github.com/basz/ember-form which shows a console message when you click any of the buttons and change the changeset...

I just noticed it happens on classic components only. If I switch over to glimmer components the described behavior does not happen...

basz:
Is this something you can work with?

snewcomer:
@basz Mind trying out the latest 3.5.7 if you don't mind and let me know if this is still an issue?

basz:
no change...

snewcomer:
I'm not sure I see an issue in this specific repo.  Glimmer components use `this.args` && `@value`.  Old reactivity paradigms like `didReceiveAttrs` and `computed` (push based system) **may** be at odds with `@tracked` and a pull based system.

The main thing here is that the `changeset` has the property you set.  If I do a bit of refactoring it looks like it does.

basz:
I don’t know enough of the internals of push and pull meganisms to make a good argument.

However I do see components are getting values set even when the underlaying value was not interacted with... In both old and glimmer components (also via render modifiers). I have a feeling because of the setProperty trigger inside the proxy. But again I am over my head here...

this[CHANGES] is completely overwritten by a new object; this[CHANGES][key] = change would feel more natural... But again even don't know if that's related at all. (tried it, didn't work :-))

would you know of a way to enable debugging on these low-level glimmer stuff, so we can see when a property is marked for an update?

Anyway... The problem I have is that I have a bunch of BsBootstrap FormElements in a form that all get value updates if one of the control changes. Not a big issue in most situations, but sometimes a control element is constructed from several other controls and their initial state is updated and rerendered. For example, I have a control that shows an image... That image is currently reloaded and results in a visual blink. Not a big deal, but still...

(no critique here on my part. I love this package)

snewcomer:
[...]

PR #517, opened 2020-06-28
Reduce proprerty notif change with tracked-built-ins

snewcomer:
ref #499


betocantu93:
Hello! got here after following a few issues about all props being notified, what's stopping this PR to be merged? 🤔  thanks!!

snewcomer:
@betocantu93 That is a great question!  I actually don't know if it will work yet.  We will have to do some testing to see how we can be more granular in our property notification updates.  Certainly a worthwhile investigation!

betocantu93:
Hello @snewcomer this PR may fix or at least its a way to fix this issue: https://github.com/poteto/ember-changeset/pull/586, its actually based on this one in the sense of avoiding replacing the `CHANGES`  entirely in favor or setDeep and instead make sure to notify nested paths

Issue #523, opened 2020-08-06
Record "isDirty=true" via [relationship].willDestory


csprocket777:
## Version
`3.8.1`

## Test Case
Model:
```
import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
export default class ProjectModel extends Model {
    @attr('string') projectName;
    @belongsTo('engagement') engagement;
    @belongsTo('collectionperiod') collectionPeriod;
    @attr('string') owner;
    @attr proxyUsers;
    @hasMany('team-criteria', {inverse: 'project'}) teams;
}
```

Template:
```
{{#with (changeset this.dataManagement.currentProject changesetKeys=this.projectChangesetKeys) as |projectChangeset|}}

...

{{/with}}
```

## Expected Behavior
Model upon load is not dirty as no changes have happened.

## Actual Behavior
`projectChangeset.changes` is reporting:
```
[
    {
        "key": "engagement.willDestroy"
    },
    {
        "key": "teams.willDestroy"
    },
    {
        "key": "_oldWillDestroy"
    },
    {
        "key": "willDestroy"
    }
]
```

snewcomer:
@csprocket777 To clarify, I see two properties we emit - `changes` && `isDirty`.  Is `isDirty` false but the changeset has `changes`?  If true, then I think I know the problem.

csprocket777:
No, `isDirty` is true and `changes` has entries in it, even though I haven't actually done anything to the data except for use it in an `ember-power-select`.

I have a dummy application representing the issue here:
https://github.com/csprocket777/changeset-powerselect-test-case

snewcomer:
https://github.com/validated-changeset/validated-changeset/pull/78

This is what we **should** be doing.  However, I don't know if this helps your case.  Doesn't seem like it would b/c `changesetKeys` relies on specificity AT or DEEPER in the path.  However, if ABOVE, it doesn't match.  And in your case, you specify `engagement` and `engagement.willDestroy` is DEEPER in the path.

csprocket777:
Is there something I should pull to give it a try?

snewcomer:
Not yet.  I'm still trying to gather ideas on how to solve this problem.  Basically we allow any level of your object tree to be mutated.  However, I didn't expect 3rd party systems to be modifying these objects (specifically ember-data objects), thus adding changes and dirtying the changeset.  The current architecture is so close to the final state I was hoping to get this library to since ~2 years ago but for this one thorny problem.  

[...]

PR #525, opened 2020-08-07
Add test for changesetKeys to helper

snewcomer:
ref #523 

@csprocket777 👋.  I'm familiar with the problem you are having.  I haven't come up with a good solution for outside factors attaching state to changesets (including opening Ember inspector).  How does this passing test compare to your setup?

### DIFF
--- a/addon/helpers/changeset.js
+++ b/addon/helpers/changeset.js
@@ -4,7 +4,7 @@ import { lookupValidator, isChangeset, isPromise, isObject } from 'validated-cha
 
 export function changeset(
   [obj, validations],
-  options= {}
+  options = {}
 ) {
   if (!obj) {
     // route transitions may trigger this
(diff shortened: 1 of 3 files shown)


csprocket777:
I think this represents my use-case right now. If i'm reading the test correctly, you'll find that it's immediately going to be `isDirty=true` once it hits the DOM (before any action taken to modify the record). Are you seeing the same or similar as I am with that test?

csprocket777:
It doesn't seem to matter how I instantiate my changeset, whether it be from the template or from JS.

csprocket777:
Ok. so if I remove the relationship keys from the `changesetKeys`, things operate as expected...

snewcomer:
@csprocket777 I'm wondering what the results are here.  Is `validationKeys` and `userChangesetKeys` what you expect?

https://github.com/validated-changeset/validated-changeset/blob/1d0e3ceb284a61e93633132865f86641c0968ebe/src/index.ts#L242

csprocket777:
What I end up seeing here is:

csprocket777:
I have an example dummy app that represents the issue I'm currently facing:
https://github.com/csprocket777/changeset-powerselect-test-case

Issue #529, opened 2020-08-14
Initial set on the empty object, when calling execute().data gives a wrong data object


RuslanZavacky:
## Version
3.8.1

## Test Case
https://github.com/poteto/ember-changeset/pull/530

## Steps to reproduce
Create a new changeset with an empty initial object, set new value, call execute and get `data`. On initial merge, data will have `value` propagated I guess from `Changes` object when merging.

```
let c = Changeset({});
c.set('org.usa.ny', 'any value');

let data = c.execute().data;
```

## Expected Behavior
`data` to be:

```
{
  org: {
    usa: {
      ny: "any value"
    }
  }
}
```

## Actual Behavior
```
{
  org: {
    usa: {
      ny: {
        value: "any value"
      }
    }
  }
}
```

RuslanZavacky:
It seems that if there is a nested object, this check is only validating single level. I guess it should be expanded to see if there is Change instance on any nested level.

<img width="534" alt="Screenshot 2020-08-14 at 13 15 21" src="https://user-images.githubusercontent.com/103556/90239483-41165300-de30-11ea-8722-f8635645af6d.png">

snewcomer:
Appreciate the issue!  `v3.8.2`

RuslanZavacky:
Thank you @snewcomer 🙏

PR #531, opened 2020-08-15
Fix nested execute() merging

snewcomer:
close #530 #529 

Minor issue in ember-release.  As of this [commit](https://github.com/emberjs/ember.js/commit/ef0e277533b3eab01e58d68b79d7e37d8b11ee34), Ember Objects have enumerable symbols on create.  We had some "nice to have" code that ensured enumerable symbols were still allowed as keys Maybe in the future we can bring this back.


RuslanZavacky:
Awesome, thank you 👏

PR #534, opened 2020-08-22
add test to demonstrate how the changeset properties are not notifying computed properties.

arenoir:
ember-changeset should be compatible with computed properties and macros.

### DIFF
--- a/tests/unit/changeset-test.js
+++ b/tests/unit/changeset-test.js
@@ -8,9 +8,10 @@ import EmberObject, {
   get,
   set,
   setProperties,
+  computed
 } from '@ember/object';
 
-import { reads } from '@ember/object/computed';
+import { reads, empty, filterBy } from '@ember/object/computed';
 import ObjectProxy from '@ember/object/proxy';
 import { dasherize } from '@ember/string';
 import { isPresent } from '@ember/utils';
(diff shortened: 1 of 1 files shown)


snewcomer:
@arenoir I think one of the major migrations our apps can make is towards a class, pull based reactivity system.  I understand that might not be feasible as well as potentially buggy for certain apps but I haven't had much luck mixing `pull` and `push` based reactivity system.  Refactoring to something like this relies on zero reactivity systems (including computed/push).  When we want to know if `hasDirtyChangesets` is true, it just calls the chain of getters.  I'm sure a more complicated example can be drummed up with reactivity + templates though.

```
    class Model extends EmberObject {
      // single
      @computed
      get changeset() {
        return Changeset(dummyModel, dummyValidator);
      }

      get isChangesetDirty() {
        return this.changeset.isDirty;
      }

      // double
      @computed
      get changesets() {
        return [Changeset(dummyModel, dummyValidator), Changeset(dummyModel, dummyValidator)];
      }

      get hasDirtyChangesets() {
        return this.dirtyChangesets.length > 0;
      }

      get dirtyChangesets() {
        return this.changesets.filter((c) => c.isDirty);
      }
    }
```

arenoir:
So what I gather you are saying is that computed properties should work within es5 classes using the `computed` decorator?   However that is not the case.  I have updated the test case to use classes and decorators.

arenoir:
I was mistaken it is working using classes and the computed decorator.

PR #546, opened 2020-09-18
merge-deep: return source when find error

### LINKED ISSUE #543: Unable to `mergeDeep` with your data. Are you trying to merge two ember-data objects?
This is more of a question, but may possibly be a bug. We recently updated to 3.9.0 and we're getting a lot of:

```
Unable to `mergeDeep` with your data. Are you trying to merge two ember-data objects?
```

Any tips on debugging what could be causing this?

snewcomer:
close #543


snewcomer:
Actually this might be a problem still.  The return will create an infinite loop as well :( since `mergeDeep` is recursive. Have to think of a different strategy.

PR #565, opened 2020-12-04
Support for validator class in validation arrays

jbescoyez:

Class validators are great! 
However, they do not work when there is more than 1 validation for a given attribute.

It would be perfect if this kind of validator map could be supported:

```
const validatorMap = {
  name: [
    validatePresence(true),
    validateLength({ min: 4 })
    new PersonalValidator()
  ],
};
```


snewcomer:
Oh interesting!  Thanks for the PR.  I'll look to fix this weekend.

jbescoyez:
Tnx @snewcomer

PR #575, opened 2021-01-28
create reproduction test

NullVoxPopuli:
Noticed this issue, but I think the problem is in validated-changeset, so I'm going to try a repro PR there as well.

The gist:
 - if a property has a key of `value`, and nested data changes, that change is not present in the `changes` array. if the key  is literally anything else and it works, which is cool -- I think I can fix this :)
 
 
 validated-changeset PR: https://github.com/validated-changeset/validated-changeset/pull/97


snewcomer:
solved downstream!

Issue #572, opened 2021-01-19
Uncaught (in promise) RangeError: Maximum call stack size exceeded at normalizeObject


mukk85:
## Version

ember-changeset: 3.10.5
ember-changeset-validations: 3.10.3

## Test Case

https://github.com/mukk85/ember-changeset-callstack-recursion-bug

## Steps to reproduce

Clone the repo above and run it with `ember s`

When using a model with a tracked property to another model normalizeObject throws a maximum call stack size exceeded error. It appears to loop infinitely inside normalizeObject.

As you see in the project, I have added a tracked property (_trade) that is assigned to a model (Company) that also has a relationship to the original model (Contact).

## Expected Behavior

It should validate.

## Actual Behavior

Throws error in title.

snewcomer:
So the problem here seems to be we get in a recursive trap with the ember-data models.  I need to figure out how to handle this appropriately since detecting "is ember-data model" isn't possible nor recommended.  We could check for the presence of keys like `_internalModel` but I know that isn't ideal as well.  Let me know if you have any ideas!

snewcomer:
v3.11.0!

PR #576, opened 2021-01-29
[Dep] bump validated-changeset 0.11.0

snewcomer:
ref https://github.com/poteto/ember-changeset/issues/572


snewcomer:
This won't solve the infinite recursion bug yet in #572 ...Still thinking and playing around with it.  Merging.

Issue #585, opened 2021-03-22
Changeset changes array is not cleared when no changes exist


jeroen-hoekstra:
## Version

3.13.2

## Test Case

[https://github.com/jeroen-hoekstra/ember-changeset-issue](https://github.com/jeroen-hoekstra/ember-changeset-issue)

## Steps to reproduce

After updating ember-changeset from **3.12.0** to **3.13.2** we noticed a difference in change tracking. Not sure if this is intended behaviour. You can run my test case with both versions and check the logs.

Open de console and notice that on page load the changeset changes array is empty. Initialy the `hidden` property is `false`. Toggle the first toggle so the `hidden` property becomes `true`. We see that the changeset changes array now contains the change object. When toggling it again the changeset changes array should become empty again, as the `hidden` property becomes equal to it's initial value, which is false. This was the case with version **3.12.0** but with **3.13.2** the changeset changes array still contains the change object, but with it's `value` set to `false`.

## Expected Behavior

**Version 3.12**

![Screenshot from 2021-03-22 09-47-11](https://user-images.githubusercontent.com/42991987/111970469-9a97dc00-8afb-11eb-987e-69b253735626.png)

## Actual Behavior

**Version 3.13.2**

![Screenshot from 2021-03-22 09-49-36](https://user-images.githubusercontent.com/42991987/111970527-a84d6180-8afb-11eb-97d9-53e73f168fe2.png)

betocantu93:
Here's a fix using #586 spike. It has a few issues to be solved over there, but... it proves(?) it fix this.

https://github.com/betocantu93/ember-changeset-issue

Issue #437, opened 2020-03-06
Modifying a buffered property of a changeset invalidates all buffered properties 


mwpastore:
Got a weird one here. Not sure if it's intentional but I suspect not. Essentially, changing any buffered property of an ember-changeset causes the observers of all buffered properties to fire (i.e. the whole buffer is invalidated). If you have an underlying object with lots of properties and associated DOM, this can get real messy, real fast. I put together a basic repro that shows (clearly, I hope) the differences in behavior between a POJO and an ember-changeset. 

## Version
3.0.6

## Test Case
https://github.com/mwpastore/-ember-changeset-notify

![](https://media.giphy.com/media/ZZlmKrS9jt0JMAjyMk/giphy.gif)

## Steps to reproduce
Modify any buffered property of a changeset. 

## Expected Behavior
Only observers of the modified property should fire. 

## Actual Behavior
The observers of all properties fire.

snewcomer:
@mwpastore Is this potentially due to the `#let` block emitting a new value.  For example, if I invert the problem - change `useChangeset = true`, toggle `Use changeset?`, both `Did Update` checkboxes are now checked.

Lmk if I am missing the goal of your repro though.

mwpastore:
@snewcomer Yes, when you switch between a POJO and a changeset, it invalidates both properties. But that's not what I'm trying to show with my repro. The difference in behavior is most easily seen when both "Did Update?" checkboxes are un-checked—i.e. by manually un-checking them one at a time; or by clicking "(clear)" to do them both at the same time. 

Maybe a "script" would help. You can run through these steps, and feel free to play with the starting conditions. It does not matter whether "Use changeset?" is initially true or false. 

When "Use changeset?" is un-checked:

1. Click "(clear)" to un-check both "Did Update?" checkboxes
1. Toggle the "Value" checkbox under Foo
1. Note that the "Did Update?" checkbox under Foo is now checked, _but checkbox under Bar remains unchecked, because the value of Bar did not change_

When "Use changeset?" is checked:

1. Click "(clear)" to un-check both "Did Update?" checkboxes
1. Toggle the "Value" checkbox under Foo
1. Note that the "Did Update?" checkbox under Foo is now checked, _and the checkbox under Bar is also now checked, even though the value of Bar did not change_

mwpastore:
[...]

PR #586, opened 2021-04-07
spike for prop notification

betocantu93:
Hello, this PR seems to fix a few bugs. (not sure if its the right way).

- #517 #585 #499 #437
It seems that notifyPropertyChange doesn't work for nested paths, so this PR runs notifyPropertyChange on all paths.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,6 +1,6 @@
 import { assert } from '@ember/debug';
 import { dependentKeyCompat } from '@ember/object/compat';
-import { BufferedChangeset } from 'validated-changeset';
+import { BufferedChangeset, Change, keyInObject } from 'validated-changeset';
 import ArrayProxy from '@ember/array/proxy';
 import ObjectProxy from '@ember/object/proxy';
 import { notifyPropertyChange } from '@ember/object';
@@ -32,8 +32,28 @@ function maybeUnwrapProxy(o) {
   return isProxy(o) ? maybeUnwrapProxy(safeGet(o, 'content')) : o;
 }
 
+function tryContent(ctx) {
+  return ctx._content ? ctx._content : ctx.content ? ctx.content : ctx;
+}
+
+function deepNotifyPropertyChange(obj, path, deep = false) {
+  var paths = path.split('.');
+  var lastPath = paths.pop(),
+    current = obj,
+    i;
+
+  for (i = 0; i < paths.length; ++i) {
+    if (current[paths[i]] == undefined) {
+    } else {
+      current = current[paths[i]];
+    }
+    notifyPropertyChange(tryContent(current), paths[i]);
+  }
+  notifyPropertyChange(tryContent(current), lastPath);
+}
+
 export class EmberChangeset extends BufferedChangeset {
-  @tracked _changes;
+  @tracked _changed;
   @tracked _errors;
   @tracked _content;
 
@@ -110,7 +130,7 @@ export class EmberChangeset extends BufferedChangeset {
   addError(key, error) {
     super.addError(key, error);
 
-    notifyPropertyChange(this, key);
+    deepNotifyPropertyChange(this, key);
     // Return passed-in `error`.
     return error;
   }
@@ -123,7 +143,7 @@ export class EmberChangeset extends BufferedChangeset {
   pushErrors(key, ...newErrors) {
     const { value, validation } = super.pushErrors(key, ...newErrors);
 
-    notifyPropertyChange(this, key);
+    deepNotifyPropertyChange(this, key);
 
     return { value, validation };
   }
@@ -133,9 +153,21 @@ export class EmberChangeset extends BufferedChangeset {
    * Returns value or error
    */
   _setProperty({ key, value, oldValue }) {
-    super._setProperty({ key, value, oldValue });
+    let changes = this[CHANGES];
 
-    notifyPropertyChange(this, key);
+    //We always set
+    this.setDeep(changes, key, new Change(value), {
+      safeSet: this.safeSet,
+    });
+
+    //If the newValue is equals to the wrapped value, delete key from changes
+    if (oldValue === value && keyInObject(changes, key)) {
+      this._deleteKey(CHANGES, key);
+    }
+
+    //Notitify deep
+    deepNotifyPropertyChange(this, key);
+    deepNotifyPropertyChange(this[CHANGES], key);
   }
 
   /**
@@ -149,7 +181,7 @@ export class EmberChangeset extends BufferedChangeset {
   _notifyVirtualProperties(keys) {
     keys = super._notifyVirtualProperties(keys);
 
-    (keys || []).forEach((key) => notifyPropertyChange(this, key));
+    (keys || []).forEach((key) => deepNotifyPropertyChange(this, key));
 
     return;
   }
@@ -159,9 +191,7 @@ export class EmberChangeset extends BufferedChangeset {
    */
   _deleteKey(objName, key = '') {
     const result = super._deleteKey(objName, key);
-
-    notifyPropertyChange(this, key);
-
+    deepNotifyPropertyChange(this, key);
     return result;
   }
 
--- a/addon/index.js
+++ b/addon/index.js
@@ -53,7 +53,7 @@ function deepNotifyPropertyChange(obj, path, deep = false) {
 }
 
 export class EmberChangeset extends BufferedChangeset {
-  @tracked _changed;
+  @tracked _changes;
   @tracked _errors;
   @tracked _content;
 


snewcomer:
I see where you are going with this and it seems reasonable!  Do you have any unanswered questions or blockers?

betocantu93:
One question: it's ok the `tryContent` approach? which looks for the obj to notify, via hardcoded .content and ._content paths, basically notifying changests and proxy alike objects?

betocantu93:
@snewcomer Dug a little bit more on this today to try to fix the tests and found out a few things...

1. `addObserver` works for deep keys
2. `notifyPropertyChange` don't work for deep keys, and thus this PR introduces `deepNotifyPropertyChange`
3. `changeset.addObserver` doesn't work as expected for a couple of reasons:

While trying to deeply notify, we loop through proxies and notify them, each of them are a *different* object from the `changeset` where clients added the observer, and thus never getting notified, this was brought to light thanks to a test.

At least with the current design, adding observers to the changeset should be added to `data` or `_content` so you can realiably add observers deeply and get notified, because the `deepNotifyPropertyChange` will always notify the `_content` or `content` of the proxies.
One downside of this is that we won't get the changeset as `this` inside the callback, but unlock adding them even with addObserver directly, `addObserver(changeset.data, 'deep.nested.key', () => {})`

```ts
  changeset.addObserver()
```

snewcomer (commented):
This PR looks like it is making progress!

addon/index.js
```diff
@@ -32,6 +32,26 @@ function maybeUnwrapProxy(o) {
...
+  var lastPath = paths.pop(),
+    current = obj,
+    i;
+
+  for (i = 0; i < paths.length; ++i) {
+    if (current[paths[i]] != undefined) {
+      current = current[paths[i]];
+    } else {
+      notifyPropertyChange(tryContent(current), paths[i]);
+    }
```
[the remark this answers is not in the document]
betocantu93: Rewrote my last comment as my understanding changed.

The idea is to traverse down to find the deepest `defined` ancestor. 

example 1.
Take: key as `foo.bar.baz` and data as `{ foo: { bar: {} } }`

The output should: `notifyPropertyChange(foo.bar, 'baz')`

example 2.
Take: key as `foo.bar.baz` and data as `{ }`, everything `undefined`

The output should: `notifyPropertyChange(data, 'foo')`

I think this could be optimized like this:
```ts
for (i = 0; i < paths.length; ++i) {
    if (current[paths[i]] != undefined) { // we can still go deep
      current = current[paths[i]];
    } else {
      //we just found that its a dynamic set, so we notify here and stop, no point in going further. 
      notifyPropertyChange(tryContent(current), paths[i]);
      return;
    }
  //We didn't stop, we haven't notified, and we went the deepest as possible, so we notify there.
  notifyPropertyChange(tryContent(current), lastPath);
}
```
I think its ok to start from the top, because we can early notify and return

betocantu93:
I'm using this with ember-m3 and seems to work great too

betocantu93:
OK, so dug a lot more today, ObjectTreeNode proxies return different references between proxy access if the prop doesn't exists inside `node.content` here: https://github.com/validated-changeset/validated-changeset/blob/d9e54bea45d0ea88d4de7e9faf39733c050c755b/src/utils/object-tree-node.ts#L39

And so we can't realiaby notify, these are the changes (need to update tests)

We'll try to find the deepest ancestor which could be notified i.e the reference to the underlaying content is the same on each access, which means deeply dynamic paths, we can't go deep or specific at all.

For now, for my bullet proof observer use case I was thinking to use a more event oriented api, to have a bullet proof solution.

In the best scenario, if the content isn't dynamic or nested, it _should_ be specific with the notifications.

i.e. 
`changeset.on('afterValidation', this.myOtherwiseObserver);`

tests/unit/changeset-test.js
```diff
@@ -1780,15 +1780,32 @@ module('Unit | Utility | changeset', function (hooks) {
...
+    let res;
+    let changeset = Changeset(dummyModel, dummyValidator);
+    changeset.data.addObserver('some.really.nested.path', function () {
+      res = changeset.get('some.really.nested.path');
+    });
+    assert.equal(undefined, changeset.get('some.really.nested.path'), 'initial value');
+    changeset.set('some.really.nested.path', 'Jack');
+    assert.equal('Jack', res, 'observer fired when setting value some.really.nested.path');
+    changeset.rollback();
+    assert.equal(undefined, res, 'observer fired with the value some.really.nested.path was rollback to');
```
snewcomer, line 1807: Can we add an assertion where `.path` is undefined?  Is that a valid test case?

tests/unit/changeset-test.js
```diff
@@ -1780,15 +1780,32 @@ module('Unit | Utility | changeset', function (hooks) {
...
+  });
 
   test('can update nested keys after rollback changes.', async function (assert) {
     let expectedResult = {
```
snewcomer, line 1811: https://github.com/poteto/ember-changeset/issues/585

Do we have a test for this issue so we can close it out when merged?

snewcomer: ping!  Lmk what you think about adding this test.  Lots of great work in this PR.

betocantu93: sorry for the delay, I ran that repo, I think it has a bug on the example itself, it uses 

```hbs
@onToggle={{action (changeset-set this.changesetObj contextKey) (not (changeset-get this.changesetObj contextKey))}}
```

I think that action helper is actually(?) mutating the CONTENT, somehow(?) and thus _it actually have changes_

but I fix I just ran this branch with a few changes to the example and its working!

```hbs
@onToggle={{changeset-set
          this.changesetObj
          contextKey
          (not (changeset-get this.changesetObj contextKey))
        }}


```

betocantu93: And so, I think... no further test need to be added? 🤔  I mean there are a few tests already checking for changes emptied after setting back the initial wrapped value? gonan double check

betocantu93:
Had to bump node because one dep needs >=12 now? it prompted because of me using `volta`, which was pin to node@10~
```ts
error ansi-regex@6.0.0: The engine "node" is incompatible with this module. Expected version ">=12". Got "10.22.0"
error Found incompatible module.
```

tests/unit/changeset-test.js
betocantu93: Hmm I ran this again and the bug still exists, but I'm unable to track _why_ it seems somehow `sets` are applied to the this[CONTENT] without executing the changeset using `changeset-set`.

package.json
```diff
@@ -75,12 +75,12 @@
     "release-it-lerna-changelog": "^1.0.3"
   },
   "engines": {
-    "node": "10.* || >= 12.*"
+    "node": ">= 12.*"
```
snewcomer, line 78: This will be a major version bump I believe.  What was the specific error?  I thought v10 had another 6 days until EOL.

addon/index.js
```diff
@@ -32,6 +32,33 @@ function maybeUnwrapProxy(o) {
...
+    lastPath = paths.pop(),
+    current = changeset,
+    i;
+
+  //If the path doesn't exists previously exist inside the CONTENT, this is a dynamic set.
+  let existsInContent = safeGet(changeset[CONTENT], path) !== undefined;
+
+  for (i = 0; i < paths.length; ++i) {
+    const curr = current[paths[i]];
+    if (existsInContent && curr && curr.content === curr.content) {
```
snewcomer, line 51: `curr` will fail for `falsey` values, correct?  

Do we want to use `getPrototypeOf` and walk the chain checking for `getOwnPropertyDescriptor`?  Or perhaps we just directly check `hasOwnProperty` on the object.

betocantu93: Yeah, I think that makes sense

betocantu93:
I noticed assigning curr=path[i] was triggering `set` on the proxy, causing trouble to fix #585 But I think we can just traverse `"content"`, or hasOwnProperty on that content as suggested... Stumbled with another bug, when you try to `changeset.get('super.deep.key')` where `content` doesn't have such structure, it fails on this line because we are trying to getDeep on undefined.

https://github.com/validated-changeset/validated-changeset/blob/d9e54bea45d0ea88d4de7e9faf39733c050c755b/src/index.ts#L963

Not sure how that should be fixed over there, any suggestion?

addon/index.js
```diff
@@ -32,31 +32,31 @@ function maybeUnwrapProxy(o) {
...
 
   //If the path doesn't exists previously exist inside the CONTENT, this is a dynamic set.
   let existsInContent = safeGet(changeset[CONTENT], path) !== undefined;
 
   for (i = 0; i < paths.length; ++i) {
-    const curr = current[paths[i]];
-    if (existsInContent && curr && curr.content === curr.content) {
-      current = current[paths[i]];
+    const curr = safeGet(getContent(current), paths[i]);
+    if (existsInContent && curr) {
```
snewcomer, line 51: Will this fail if `curr` is falsey?

snewcomer:
@betocantu93 Yes a fix over in validated-changeset with a simple test case seems prudent!  Would happily accept.

addon/index.js
snewcomer: ping!

addon/index.js
```diff
@@ -32,6 +32,33 @@ function maybeUnwrapProxy(o) {
   return isProxy(o) ? maybeUnwrapProxy(safeGet(o, 'content')) : o;
 }
 
+function getContent(ctx) {
+  return ctx._content ? ctx._content : ctx.content ? ctx.content : ctx;
```
snewcomer, line 36: Is `_content` meant to capture from the Changeset and `content` from an ObjectProxyNode?

betocantu93: Yes, that's correct.

betocantu93:
Two more issues on my mind:
1.
> Stumbled with another bug, when you try to changeset.get('super.deep.key') where content doesn't have such structure, it fails on this line because we are trying to getDeep on undefined.
> https://github.com/validated-changeset/validated-changeset/blob/d9e54bea45d0ea88d4de7e9faf39733c050c755b/src/index.ts#L963

Not exactly sure how to handle that case inside the `get` function

2. I also noticed an inconvenience. As we are notifying the [CONTENT] well, it will trigger all the observers in the wrapped model, which is ok for most scenarios, i.e a custom route to edit a model, but if by any UX you have both the model and the changeset rendered (think of a table with model rows on the left, with a flyout changeset editing) it will trigger whatever observers you might have on the changeset _and_ the model row, so you might have unexpected UX... 

What do you think about the second point? would it make this approach non ideal?

My counter proposal is to have a tracked mirror/fake structure that we will be notified and setted just for observavility purposes, instead of the content. We just have to be sure (_I think_ 🤔) that we consume tags per get/set, with that octane getters should work and classic computeds _should_ work, and for `addObserver`, we might have to advise to hook em to this mirror... i.e. `Ember.addObserver(changeset._mirror, 'some.nested.prop');

snewcomer:
Fixing 1 in #590!  Thanks!

Regarding 2, it sounds like we have to get really really granular on our KVO notifications.  Just for some clarification, is the problem solely with `addObserver`?  Or are there other cases that don't work. 

Overall, we do need to get this right as you have pointed out - `notifyPropertyChange` just doesn't work right now with nested paths (or perhaps does but fires to broadly)

betocantu93:
I was thinking in something like this in pseudocode: 

```ts
//ember-changeset
new Proxy(obj, {
   get(target, path) {
       Ember.get(target[MIRROR], path); //how to handle dynamic structures? can we setDeep before consuming?
       .... continue default
   },

  set(target, path, value) {
      setDeep(target[MIRROR], path, value //I think we don't care about the value);
      Ember.get(target[MIRROR], path); // Do we need to consume?
      ....continue default
  }
})

```

betocantu93:
How does the `changeset-get` helper gets to recompute 🤔

betocantu93:
The problem with native getters and tracked is that every consumed tag is a dependency, for example:

Screen Shot 2021-05-09 at 18 00 35

addon/index.js
snewcomer: Sry for stringing this along but I do want to get a solution in.  I'm learning from your work to understand the issue better.

So is it best to say, not only does notifyPropertyChange not work for nested keys (this is obvious to me), but also we are notifying on the wrong object?  Meaning `notifyPropertyChange(this, key)` leads to the wrong/no expected behaviour as well?  I wouldn't have through this would have work for an ObjectProxyNode (non Ember object, etc)

PR #591, opened 2021-05-10
Try mirror

betocantu93:
@snewcomer This is another approach for #586, property notifications

We use tracked-built-ins to have a common tracked object reference that we can keep notifying of changes by lazily consuming tags on every changeset.get.

This PR also implements `addObserver` and `removeObserver` to allow client code to add / remove observers in an ergonomic way to a changeset.

The only "downside" AFAIK is that for classic computeds, we have to manually subscribe to the `changeset.mirror` property. One way to solve this, is to just use native getters or, create a native getter which consumes the value and then add that native getter to the dependencies of your computed, i.e.

```ts
get myProp() {
  return this.args.changeset.get(this.args.path);
}

@computed('myProp')
get someComputed() {
 return ....
}

//or
@computed(`args.changeset.mirror.yourpath`)
...

```

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,6 +1,6 @@
 import { assert } from '@ember/debug';
 import { dependentKeyCompat } from '@ember/object/compat';
-import { BufferedChangeset } from 'validated-changeset';
+import { BufferedChangeset, Change, keyInObject } from 'validated-changeset';
 import ArrayProxy from '@ember/array/proxy';
 import ObjectProxy from '@ember/object/proxy';
 import { notifyPropertyChange } from '@ember/object';
@@ -32,8 +32,28 @@ function maybeUnwrapProxy(o) {
   return isProxy(o) ? maybeUnwrapProxy(safeGet(o, 'content')) : o;
 }
 
+function tryContent(ctx) {
+  return ctx._content ? ctx._content : ctx.content ? ctx.content : ctx;
+}
+
+function deepNotifyPropertyChange(obj, path, deep = false) {
+  var paths = path.split('.');
+  var lastPath = paths.pop(),
+    current = obj,
+    i;
+
+  for (i = 0; i < paths.length; ++i) {
+    if (current[paths[i]] == undefined) {
+    } else {
+      current = current[paths[i]];
+    }
+    notifyPropertyChange(tryContent(current), paths[i]);
+  }
+  notifyPropertyChange(tryContent(current), lastPath);
+}
+
 export class EmberChangeset extends BufferedChangeset {
-  @tracked _changes;
+  @tracked _changed;
   @tracked _errors;
   @tracked _content;
 
@@ -110,7 +130,7 @@ export class EmberChangeset extends BufferedChangeset {
   addError(key, error) {
     super.addError(key, error);
 
-    notifyPropertyChange(this, key);
+    deepNotifyPropertyChange(this, key);
     // Return passed-in `error`.
     return error;
   }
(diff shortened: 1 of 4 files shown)


addon/index.js
```diff
@@ -188,6 +280,26 @@ export class EmberChangeset extends BufferedChangeset {
 
     return this;
   }
+
+  get(key) {
+    safeGet(this[VIRTUAL_MIRROR], key); //consume tag for native tracked dependencies
+    return super.get(...arguments);
+  }
```
snewcomer, line 287: Does this handle the `obj.key` case as well?  Or do we have the ability to handle this case?

addon/index.js
```diff
@@ -56,6 +128,11 @@ export class EmberChangeset extends BufferedChangeset {
     return safeSet(obj, key, value);
   }
 
+  @dependentKeyCompat
+  get mirror() {
+    return this[VIRTUAL_MIRROR];
+  }
```
snewcomer, line 134: Is this meant to be access anywhere?  I don't see `this.mirror` anywhere.

addon/index.js
```diff
@@ -32,10 +34,80 @@ function maybeUnwrapProxy(o) {
...
+  for (i = 0; i < paths.length; ++i) {
+    if (Object.prototype.hasOwnProperty.call(current, paths[i])) {
+      current = safeGet(current, paths[i]);
+    } else {
+      maybeDynamicPathToNotify = paths[i];
+      break;
+    }
+  }
+  const pathToNotify = maybeDynamicPathToNotify ? maybeDynamicPathToNotify : lastPath;
+  notifyPropertyChange(current, pathToNotify);
```
snewcomer, line 53: I wonder if we should make `notifyPropertyChange` opt-in in a next major 4.0 release?  Generally, we gave patterns that obfuscate the need for these reactivity systems.

snewcomer: Or am I missing a crucial need for `notifyPropertyChange`, even in the new reactivity world?

addon/index.js
betocantu93: Mainly for public access, for the classic computeds case

addon/index.js
```diff
@@ -188,6 +280,26 @@ export class EmberChangeset extends BufferedChangeset {
...
+    return super.get(...arguments);
+  }
+
+  addObserver(key, callback) {
+    if (!deepHasOwnProperty(this[VIRTUAL_MIRROR], key)) {
+      this.setDeep(this[VIRTUAL_MIRROR], key);
+    }
+
+    deepAddObserver(this[VIRTUAL_MIRROR], key, callback);
+  }
```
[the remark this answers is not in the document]
betocantu93: Manually adding observers in general is not that common, I needed them for a very specific use case that I'm not entirely sure if it could be solved by native getters and this mirror approach, still having a way to add them gives more flexibility and an actual way to do it, which doesn't exist today, also... adding them to the `content` have it's own set of inconveniences

addon/index.js
betocantu93: 🤔🤔🤔 maybe it's no longer needed, you're right, gonna have a look, one think I still need to look for is reactivity around arrays, i.e. computeds which use `changeset.someArray.@each.value`

addon/index.js
```diff
@@ -188,6 +280,26 @@ export class EmberChangeset extends BufferedChangeset {
 
     return this;
   }
+
+  get(key) {
+    safeGet(this[VIRTUAL_MIRROR], key); //consume tag for native tracked dependencies
```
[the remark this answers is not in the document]
betocantu93: I followed the pattern of using this[KEYWORD] on internal usages, the get mirror(){} is meant to be used by client code, public access, like.... get data(){}

addon/index.js
betocantu93: I think yes? Because of the Proxy traps all access 🤔

Issue #594, opened 2021-05-17
Different forms of 'get' return different results for ember-data models


flynnawtc:
## Version
main

## Test Case

Thank you @snewcomer for the prompt response and fix to #592 
I have been testing it since you updated validated-changeset earlier, and it does resolve the problem in most cases, but when I am working with ember-data objects in (ember 3.24) I am still seeing what look like problems, but you may be able to tell me that I'm doing it wrong.

See test project uploaded as https://github.com/flynnawtc/cs-test

The behaviour of the changeset appears to differ depending on whether an Ember-Data object is used or not.

When I construct a changeset as:
```Javascript
    let trunk = { branch: { leaf: null } };
    let leaf1 = {
      id: 1,
      description: 'leaf 1',
    };
    let cs = Changeset(trunk, null);
    cs.set('branch.leaf', leaf1);
```
then 
I can use any of the following to get the same result ('leaf 1'):
```Javascript
cs.get('branch.leaf.description')
cs.get('branch.leaf').description
cs.get('branch.leaf').get('description')
```

However when I construct the changeset using ember-data:
```Javascript
    let trunk = this.store.createRecord('trunk');
    let branch = this.store.createRecord('branch');
    let leaf1 = this.store.createRecord('leaf', {
      id: 1,
      description: 'leaf 1',
    });
    trunk.branch = branch;
    let cs = Changeset(trunk, null);
    cs.set('branch.leaf', leaf1);
```
then this works, returning 'leaf 1':
```Javascript
cs.get('branch.leaf.description')
```
but these don't - they return 'undefined'
```Javascript
cs.get('branch.leaf').description
cs.get('branch.leaf').get('description')
```

I have tried constructing this as a test, but failed, hence the project referenced above.
Thanks again for your efforts which are much appreciated.
Regards
Adrian

snewcomer:
@flynnawtc I think I might have your fix here.

https://github.com/validated-changeset/validated-changeset/pull/120

Issue #592, opened 2021-05-16
#set with Object should work repeatedly for nested properties


flynnawtc:
## Version

master

## Test Case

```Javascript
  test('#set Ember.set with Object actually does work TWICE for nested', async function (assert) {
    set(dummyModel, 'name', {});
    let title1 = { id: 'Mr', description: 'Mister' };
    let title2 = { id: 'Mrs', description: 'Missus' };
    let dummyChangeset = Changeset(dummyModel);
    set(dummyChangeset, 'name.title', title1);

    assert.equal(get(dummyModel, 'name.title.id'), undefined, 'should not have new change');
    assert.equal(dummyChangeset.name.title.id, 'Mr', 'should have new change');
    assert.equal(dummyChangeset.get('name.title.id'), 'Mr', 'should have new change using get');

    let changes = get(dummyChangeset, 'changes');
    assert.deepEqual(changes, [{ key: 'name.title', value: title1 }], 'changes with nested key Ember.set');
    //This is expected to fail but is retained to show us what is stored in changes
    assert.deepEqual(changes, [{ key: 'name.title.id', value: 'Mr' }], 'changes with nested key Ember.set');

    set(dummyChangeset, 'name.title', title2);

    assert.equal(get(dummyModel, 'name.title.id'), undefined, 'should not have new change');
    assert.equal(dummyChangeset.name.title.id, 'Mrs', 'should have new change');
    assert.equal(dummyChangeset.get('name.title.id'), 'Mrs', 'should have new change using get');

    changes = get(dummyChangeset, 'changes');
    assert.deepEqual(changes, [{ key: 'name.title', value: title2 }], 'changes with nested key Ember.set');
    //This is expected to fail but is retained to show us what is stored in changes
    assert.deepEqual(changes, [{ key: 'name.title.id', value: 'Mrs' }], 'changes with nested key Ember.set');

    dummyChangeset.execute();

    assert.equal(dummyModel.name.title.id, 'Mrs', 'has new property');
  });

  test('#set with Object should work TWICE for nested', async function (assert) {
    set(dummyModel, 'name', {});
    let title1 = { id: 'Mr', description: 'Mister' };
    let title2 = { id: 'Mrs', description: 'Missus' };
    let dummyChangeset = Changeset(dummyModel);
    dummyChangeset.set('name.title', title1);

    assert.equal(get(dummyModel, 'name.title.id'), undefined, 'should not have new change');
[...]

PR #595, opened 2021-05-17
[Test]: ember-data nested set

snewcomer:
ref https://github.com/poteto/ember-changeset/issues/592

ref https://github.com/poteto/ember-changeset/issues/594

### DIFF
--- a/tests/unit/changeset-test.js
+++ b/tests/unit/changeset-test.js
@@ -881,7 +881,7 @@ module('Unit | Utility | changeset', function (hooks) {
     assert.equal(dummyChangeset.name.title.id, 'Mr', 'should have new change');
     assert.equal(dummyChangeset.get('name.title.id'), 'Mr', 'should have new change using get');
 
-    let changes = get(dummyChangeset, 'changes');
+    let changes = dummyChangeset.changes;
     assert.deepEqual(changes, [{ key: 'name.title', value: title1 }], 'changes with nested key Ember.set');
 
     set(dummyChangeset, 'name.title', title2);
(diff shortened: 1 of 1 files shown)


snewcomer:
@flynnawtc Let me know how I can improve this test to make it fail!

flynnawtc:
Hi @snewcomer 
I really appreciate your taking an interest in this.
Here is an updated test that tries all the various means of accessing the properties of the nested object. Some of these fail.
My motivation for following this up, is to use ember-power-select on forms using ember-data models and changesets.

```Javascript
  test('#set Ember.set with Ember Data Object actually does work TWICE for nested', async function (assert) {
    let store = this.owner.lookup('service:store');

    let mockProfileModel = store.createRecord('profile');
    let mockUserModel = store.createRecord('user', {
      profile: mockProfileModel,
      save: function () {
        return Promise.resolve(this);
      },
    });

    let dummyChangeset = Changeset(mockUserModel);
    let pet1 = store.createRecord('dog', { breed: 'jazzy' });
    let pet2 = store.createRecord('dog', { breed: 'hands' });

    set(dummyChangeset, 'profile.pet', pet1);

    assert.equal(dummyChangeset.profile.pet.breed, 'jazzy', 'should have new change');
    assert.equal(dummyChangeset.get('profile.pet.breed'), 'jazzy', 'should have new change using get');
    assert.equal(dummyChangeset.get('profile.pet').breed, 'jazzy', 'should have new change using get-object.property');
    assert.equal(dummyChangeset.get('profile.pet').get('breed'), 'jazzy', 'should have new change using get-object.get-property');
    let p1 = dummyChangeset.get('profile.pet');
    assert.ok(p1 !== null);
    assert.equal(p1.breed, 'jazzy', 'got object should have new change');
    assert.equal(p1.get('breed'), 'jazzy', 'got object should have new change using get-property');

    p1 = dummyChangeset.profile.pet;
    assert.ok(p1 !== null);
    assert.equal(p1.breed, 'jazzy', 'object should have new change');
    assert.equal(p1.get('breed'), 'jazzy', 'object should have new change using get-property');


    let changes = dummyChangeset.changes;
    assert.equal(changes[0].value.breed, 'jazzy', 'changes with nested key Ember.set');

    set(dummyChangeset, 'profile.pet', pet2);

    assert.equal(dummyChangeset.profile.pet.breed, 'hands', 'should have new change');
    assert.equal(dummyChangeset.get('profile.pet.breed'), 'hands', 'should have new change using get');
    assert.equal(dummyChangeset.get('profile.pet').breed, 'hands', 'should have new change using get-object.property');
    assert.equal(dummyChangeset.get('profile.pet').get('breed'), 'hands', 'should have new change using get-object.get-property');
    let p2 = dummyChangeset.get('profile.pet');
    assert.ok(p2 !== null);
    assert.equal(p2.breed, 'hands', 'got object should have new change');
    assert.equal(p2.get('breed'), 'hands', 'got object should have new change using get-property');

    p2 = dummyChangeset.profile.pet;
    assert.ok(p2 !== null);
    assert.equal(p2.breed, 'hands', 'object should have new change');
    assert.equal(p2.get('breed'), 'hands', 'object should have new change using get-property');

    changes = dummyChangeset.changes;
    assert.equal(changes[0].value.breed, 'hands', 'changes with nested key Ember.set');

    dummyChangeset.execute();

    assert.equal(get(mockUserModel, 'profile.pet.breed'), 'hands', 'has new property');
  });
```
Hope this makes sense.
Adrian

snewcomer:
You are right.  I'm thinking there are two issues at play.  

1. async vs sync relationships.  I imagine the latter is easier.
2. properties on an ember data model return `hasOwnProperty('breed') === false`.   See [here](https://github.com/validated-changeset/validated-changeset/blob/1c904b9524565d417834f8e19a7085432ba0f9cf/src/utils/object-tree-node.ts#L21) for reference.  

The `get` with the full path should work every time.  But the cases you have laid out `.get('profile.pet').breed` might not.  

To fix 2, I wonder if we should try accessing the property instead of checking hasOwnProperty 🤔.  I don't remember why I did it that way.

flynnawtc:
Because we use RESTAdapter rather than JSONAPIAdapter to suit our back-end, we have never needed to consider sync/async as we retrieve 'fully populated' object models from the server each time.

I'm afraid you're gone beyond me when you delve in to the workings of an ember-data model. It's something I have never really looked at.

The problems have arisen when building forms with a selection box that, say, allows a user to choose a pet for the user.
The user model becomes the basis for the changeset.
The list of pets is passed to ember-power-select for display to the user.
=> we are dealing with nested properties (profile.pet)

I have added some sample usages to my test project https://github.com/flynnawtc/cs-test if you have a moment to take a look.

The documented recommendation to use changeset-get for nested properties does not work, while Ember-get seems to work for ember data models. When using regular Javascript objects, then both methods work.
This was my initial starting point which resulted in the tests above.

In my production code I have now worked around the issue as per testselect2 in the example but it would probably be worthwhile to figure out why testselect is failing.

PR #600, opened 2021-06-10
Rework final phases of get trap

snewcomer:
prioritize internal data over Proxy instance getters/methods

ref https://github.com/validated-changeset/validated-changeset/pull/123


snewcomer:
need to rethink this

PR #627, opened 2021-12-07
[WIP] Rewrite to use native Proxy

BryanCrotaz:
Work in progress. Using existing test suite plus some more edge cases and preserving public API.

Enables developer to use normal JS semantics:

```js
var output = changeset.gruffalos.lastObject.say('grrr');
changeset.people[1].name.first = 'Bryan';
```

Really useful for having nested components:
```hbs
{{#let (changeset this.model) as |changeset|}}
  {{#each changeset.people as | person |}}
    <PersonForm @model={{person}}/>
  {{/each}}
{{/let}}
```
PersonForm doesn't have to know that this is a changeset. The save and rollback can be at a higher level.


addon/classes/ChangesetArrayProxyHandler.ts
```diff
@@ -0,0 +1,298 @@
...
+import IChangesetProxyHandler from './IChangesetProxyHandler';
+import ProxyOptions from './ProxyOptions';
+import { ProxyArrayValueKey } from 'ember-changeset/utils/symbols';
+
+type Input = any[] | ArrayProxy<any>;
+
+export default class ChangesetArrayProxyHandler<T extends Input> implements IChangesetProxyHandler<T> {
+  constructor(source: T, _options?: ProxyOptions) {
+    if (source?.constructor?.name === 'ArrayProxy') {
+      this.__sourceProxy = source as ArrayProxy<any>;
```
NullVoxPopuli, line 14: these should either by symbol properties, or private properties, as they may conflict with folks data

addon/classes/ChangesetArrayProxyHandler.ts
```diff
@@ -0,0 +1,298 @@
...
+      this.__source = (source as ArrayProxy<any>).toArray();
+    } else {
+      // wrap the array so it's tracked
+      this.__source = A(source as []);
+    }
+  }
+
+  public get(_target: T, key: string): any {
+    // extra stuff
+    switch (key) {
```
NullVoxPopuli, line 24: you can shorten this by doing `if (typeof this.readArray[key] === 'function') return this.readArray[key];`, etc

addon/classes/ChangesetArrayProxyHandler.ts
```diff
@@ -0,0 +1,298 @@
...
+      return this.__nestedProxies.get(index);
+    } else {
+      return this.readArray[index];
+    }
+  }
+
+  @action
+  public getValue(key: string | symbol) {
+    switch (key) {
+      // backwards compatible
```
NullVoxPopuli, line 126: yay back compat

addon/utils/is-unchanged.ts
```diff
@@ -0,0 +1,7 @@
+export default function isUnchanged(a: any, b: any) {
```
NullVoxPopuli, line 1: I'm loving all these utilities

NullVoxPopuli:
So, I like where this is going, but the real implementation is here: https://github.com/validated-changeset/validated-changeset
is there a reason you opted to only update this library? or is there a PR I'm missing? :thinking:

addon/classes/ChangesetArrayProxyHandler.ts
BryanCrotaz: you mean __sourceProxy etc?

addon/classes/ChangesetArrayProxyHandler.ts
BryanCrotaz: If you notice I have two different ways of doing it
readArray[key] returns the proxy or the source if there's no changes
writeArray[key] creates a proxy if it doesn't exist and returns it

I thought about using a map to choose which one to return. Otherwise `readKeys.includes[key]` etc

Thoughts?

addon/classes/ChangesetArrayProxyHandler.ts
BryanCrotaz: I don't think these do conflict. Proxy is not a wrapper that falls back to the internals like Ember.ObjectProxy. It explicitly calls in with `get` and `set` and you do your own fallback. __sourceProxy is private api so it's not available through get and set, and if there was a name conflict it would fall through to the proxied object correctly.

NullVoxPopuli:
> validated-changeset is supposed to be independent of Ember

is true -- but it's also where _most_ of the complexity comes in.

is the goal to eliminate the need for validated-changeset?

and fwiw, I think it _is_ possible to a better native proxy design with validated-changeset _and_ provide deep reactivity as well

BryanCrotaz:
My goals are to have ember-changes et working for deep nesting without get and set. If that means submitting to validated- changeset instead then that's fine by me

NullVoxPopuli:
Ah ok, we gan probably do even less work then. Thanks for clarifying!

addon/classes/ChangesetArrayProxyHandler.ts
BryanCrotaz: `typeof this.readArray[key] === 'function'` doesn't work - typescript expects the index to be a number. Is there a way to iterate methods on an array without [key]?

NullVoxPopuli: I aw not sure

addon/classes/ChangesetArrayProxyHandler.ts
```diff
@@ -0,0 +1,298 @@
...
+    // or if a nested proxy is dirty
+    return (this.__proxyArray !== undefined);
+  }
+
+  public get isPristine(): boolean {
+    return (this.__proxyArray === undefined);
+  }
+
+  public get isValid(): boolean {
+    // TODO: validate this level
```
NullVoxPopuli, line 94: Todo!

addon/classes/ChangesetArrayProxyHandler.ts
```diff
@@ -0,0 +1,298 @@
...
+
+    // now look at all the nested proxies
+    for (var proxy of this.__nestedProxies.values()) {
```
NullVoxPopuli, line 97: Why reimplement these behaviors? Couldn't you return tte value from the underlying validated-changeset?

addon/classes/ChangesetObjectProxyHandler.ts
```diff
@@ -0,0 +1,394 @@
...
+      case 'data':
+      case 'isChangeset':
+      case 'isDirty':
+      case 'isPristine':
+      case 'isValid':
+      case 'pendingData':
+        throw `changeset.${key} is readonly`;
+    }
+    // nested keys are separated by dots
+    var [localKey, subkey] = splitKey(key as string);
```
NullVoxPopuli, line 184: What about deeply nested keys? Maybe 5+ layers deep?

addon/classes/ChangesetObjectProxyHandler.ts
```diff
@@ -0,0 +1,394 @@
...
+      }
+      return true;
+    }
+  }
+
+  private markChange(localKey: string, value: any) {
+    // have to use get() here because source might be an EmberProxy
+    const oldValue = get(this.__source, localKey);
+    const unchanged = isUnchanged(value, oldValue);
+    let changes = this.__changes;
```
NullVoxPopuli, line 222: Wait, does this PR remove and re-implement validated-changeset?

Sorry if obvious

BryanCrotaz:
closing in favour of https://github.com/validated-changeset/validated-changeset/pull/150

Issue #629, opened 2022-01-18
moment converts to proxy on changeset creation


bgolla:
## Version 

    "ember-changeset": "^4.0.0-beta.1",
    "ember-changeset-validations": "4.0.0-beta.2",

## Steps to reproduce
    let x = new Changeset({ foo: moment() });
    console.log(x.foo instanceof moment); // returns false
    console.log(x.foo); //returns Proxy {changes: {…}, content: Moment, proxy: Proxy, safeGet: ƒ, isObject: ƒ, …}

## Expected Behavior
Returns moment instance instead 

## Actual Behavior
Returns Proxy

snewcomer:
@bgolla Thank you!  Is there a minimal test and failing assertion (using this repository's test suite) that shows this bug?

snewcomer:
@bgolla https://github.com/poteto/ember-changeset/pull/636#issuecomment-1064763004

snewcomer:
https://github.com/poteto/ember-changeset/issues/628#issuecomment-1065797210

veelci:
@snewcomer Why did you close this? Nothing has been done to fix this issue?

PR #636, opened 2022-02-11
instanceof test

luxzeitlos:
This is for now just a failing test, and I'm not even sure this is an error of this repo.
The thing is, I'm not even sure this is a real problem and where this must be fixed.

However I'm very confident this is the error described in #629. I encountered the same Problem with a `Temporal`, which basically completely breaks. Utilizing `.content` works, however this is not nice.

The current behaviour is:

```js

    class Foo {}
    let model = EmberObject.create({ foo: new Foo() });
    let dummyChangeset = Changeset(model);

    // this *does* work!
    assert.ok(dummyChangeset.foo.content instanceof Foo, 'content property is an instance');
    
    // this *does not* work!
    assert.ok(dummyChangeset.foo instanceof Foo, 'proxy itself is an instance');
```

### DIFF
--- a/tests/unit/changeset-test.js
+++ b/tests/unit/changeset-test.js
@@ -2754,4 +2754,13 @@ module('Unit | Utility | changeset', function (hooks) {
     changeset.unexecute();
     assert.ok(true); // we just want no error until here
   });
+
+  test('instanceof checks on custom classes work`', async function (assert) {
+    class Foo {}
+    let model = EmberObject.create({ foo: new Foo() });
+    let dummyChangeset = Changeset(model);
+
+    assert.ok(dummyChangeset.foo.content instanceof Foo, 'content property is an instance');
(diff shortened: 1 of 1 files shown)


snewcomer:
@luxferresum @bgolla so the problem here is the intermediate value is the proxy.  The assumption was that a user only ever wants the leaf key to display in the UI.  So in your example,

```
class Foo { name = 'Scott'; } 
```

`changeset.foo` will be the proxy and `changeset.foo.name === 'Scott'`.  

What concrete problem are you running up against?

veelci:
{ beginDate:

veelci:
```
const beginDate = moment('2020-04-10');
const params = { beginDate: beginDate.clone() };
const changeset = Changeset(params);

// This does not work
assert.true(changeset.beginDate.isSame(today))
```

PR #646, opened 2022-03-21
Bump validated-changeset 1.2.0

sapryniukt:


package.json
```diff
@@ -42,7 +42,7 @@
     "@glimmer/tracking": "^1.0.4",
     "ember-auto-import": "^2.3.0",
     "ember-cli-babel": "^7.26.6",
-    "validated-changeset": "~1.1.2"
+    "validated-changeset": "~1.2.0"
```
snewcomer, line 45: this significantly changed the package-lock file.  Should be limited changes.  Probably npm install isn't respecting the Volta version in package.json

snewcomer: how its right.  It wasn't before.  It was bumped to lock file version 2, which would be node 14 or 16

PR #660, opened 2022-10-19
Demonstrate dirty change tracking failure

NullVoxPopuli:
Editing one value should not dirty other values.

This can be seen by editing any of the fields in the paused test, and seeing two logs print.
When this bug is fixed, only the log for the changed field should print.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,16 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@ampproject/remapping": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz",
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
https://github.com/NullVoxPopuli/ember-deep-tracked/blob/main/ember-deep-tracked already solved this via Proxies :D

sukima:
I realized that dirty changes on a global standpoint is a big issue for asynchronous validations! If you have ten async validators and you make a change to _one_ field all ten will spawn network requests!

sukima:
As an example on how best to approach the issue of deep tracking [Immer.js](https://immerjs.github.io/immer/) is a prime example of tracking changes via on-demand proxies.

NullVoxPopuli:
> As an example on how best to approach the issue of deep tracking [Immer.js](https://immerjs.github.io/immer/) is a prime example of tracking changes via on-demand proxies.

It's there any advantage to this for our purposes over deep-tracked? 🤔

sukima:
> It's there any advantage to this for our purposes over deep-tracked? 🤔

A super quick scan of your ember-deep-tracked and it looks like it is doing as I described above. It looks like it is generating a proxy for each level on demand (when the getter/setter is called). Pretty sure Immur does this also.

PR #662, opened 2023-01-10
Pass through original mergeDeep error in cause property

mintyfloss:

This PR passes through the original error with `mergeTargetAndSource` to help with debugging. I'm getting the `mergeDeep` error, but I don't have enough info to debug. I recently noticed that there's a `cause` property that can be used to pass through errors. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause.


SergeAstapov:
Thank you @mintyfloss !

PR #666, opened 2023-04-14
Use #let instead of deprecated #with in example

x-m-el:
Apperently `{{#with}}` is deprecated and `{{#let}}` should be used instead. I changed this in the examples to avoid confusion.

Note: I got this info from Ember discord, I don't have an official place where this is mentioned.


SergeAstapov:
Thank you @x-m-el !

Issue #361, opened 2019-08-02
 A way to remove errors without rolling back


AleksandrBorovkov:
Is there some way to clear errors without rollback in changeset?  I've faced the same issue as here https://github.com/poteto/ember-changeset/issues/282

snewcomer:
👋 What version are you on? I'm guessing `rollbackProperty`/`rollbackInvalid` doesn't work for you?

https://github.com/poteto/ember-changeset/pull/307

theroncross:
I would also like to see this added. Use-case is as follows:

I have a fieldset of checkboxes, with at least one required, and each value included in the changeset. User deselects all checkboxes. On the last checkbox, the validation adds an error for that key. Now user re-selects another checkbox. Validation passes, but doesn't remove the error.

MichalBryxi:
Chiming in with our use-case: We do a lot of manual stuff, because ... reasons. At one point we figure out that validation errors on all the fields are not true anymore and should be cleared. Not the values, just the errors.
I don't think anything from the existing API covers that.

MichalBryxi:
I _think_ this trick should work around the problem before we have proper support:

```js
for (const errKey in this.changeset._errors) {
  this.changeset.addError(errKey); // no second argument effectively clears the error
}
```

PR #668, opened 2023-05-17
add #removeError and #removeErrors to Changeset interface

jkeen:
I'm using ember-changeset with ember-changeset-validations with some validations where the presence requirement of one field is dependent on the value of another. If the dependent field ends up with an error and then the field it depends on changes, the error should really be cleared. But it didn't seem like there was a clean way to do that without actually rolling back the property. This seems to have come up a number of times in issues and PRs.

> This seems to be something that works in the newer `validated-changeset` 4.x-5.x experiment, but is not something available in the classic BufferedChangeset. 

> https://github.com/poteto/ember-changeset/issues/361
> https://github.com/poteto/ember-changeset/issues/282

> There were some workarounds in the above threads, like effectively "clearing" an error by changeset.addError('email', null), but in my testing it still left the changeset with an isValid = false state.

This PR adds removeError and removeErrors to the Changeset interface, which was just added to validated-changeset 1.4.0

### DIFF
--- a/ember-changeset/addon/index.js
+++ b/ember-changeset/addon/index.js
@@ -145,7 +145,6 @@ export class EmberChangeset extends BufferedChangeset {
     super.removeError(key);
 
     notifyPropertyChange(this, key);
-    // Return passed-in `error`.
     return this;
   }
 
--- a/ember-changeset/addon/index.js
+++ b/ember-changeset/addon/index.js
@@ -137,6 +137,29 @@ export class EmberChangeset extends BufferedChangeset {
   }
 
   /**
+   * Manually remove an error from the changeset.
+   *
+   * @method removeError
+   */
+  removeError(key) {
+    super.removeError(key);
+
+    notifyPropertyChange(this, key);
+    // Return passed-in `error`.
+    return this;
+  }
+
+  /**
+   * Manually clears the errors from the changeset
+   *
+   * @method removeError
+   */
+  removeErrors() {
+    super.removeErrors();
+    return this;
+  }
+
+  /**
    * Manually push multiple errors to the changeset as an array.
    *
    * @method pushErrors
--- a/test-app/tests/unit/changeset-test.js
+++ b/test-app/tests/unit/changeset-test.js
@@ -3100,6 +3100,94 @@ module('Unit | Utility | changeset', function (hooks) {
   });
 
   /**
+   * #removeError
+   */
+
+  test('#removeError removes an error from the changeset', async function (assert) {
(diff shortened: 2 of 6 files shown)


jkeen:
@SergeAstapov The tests will pass on this branch now that `validated-changeset` 1.4.0 has been released. Think we could get this merged and released too?

ember-changeset/addon/index.js
```diff
@@ -136,6 +136,29 @@ export class EmberChangeset extends BufferedChangeset {
...
+  /**
+   * Manually remove an error from the changeset.
+   *
+   * @method removeError
+   */
+  removeError(key) {
+    super.removeError(key);
+
+    notifyPropertyChange(this, key);
+    // Return passed-in `error`.
```
SergeAstapov, line 148: Should we update this comment?

test-app/tests/unit/changeset-test.js
```diff
@@ -3099,6 +3099,124 @@ module('Unit | Utility | changeset', function (hooks) {
     assert.ok(get(dummyChangeset, 'isValid'), 'should be valid');
   });
 
+  /**
+   * #removeError
+   */
+
+  test('#removeError removes an error from the changeset', async function (assert) {
```
SergeAstapov, line 3106: @jkeen could you please either add new test or add an assertion to this one to that checks that field values does not get reset? I assume it’s an important aspect of this new method

SergeAstapov (commented):
Thank you @jkeen for contribution! Do you mind to address couple of comments to polish this and then we can land this.

test-app/tests/unit/changeset-test.js
jkeen: Great point, added a new test to verify that @SergeAstapov

jkeen:
@SergeAstapov Anything else holding this up from being merged and a new release cut? I'd love to get my production app off my hacky pnpm patches

SergeAstapov:
@jkeen this is merged. we need to solve npm access/npm token issue before this could be released.

jkeen:
@SergeAstapov Oohhh is that why the releases have stalled? I saw that #689 hadn't been merged in a while and was wondering

It took me a year to get a token sorted for a ruby gem I took over maintenance for, so I feel your pain and appreciate your efforts!

jkeen:
@SergeAstapov Sorry to bug ya again about this, but did you get that NPM token worked out? 😃

ryan-nauman-cs:
could a release be published with this? i just went down a rabbit hole of trying to use it before its available 😆

PR #691, opened 2024-10-23
Modernize readme examples

SergeAstapov:
Readme has a few pre-Octane examples and patterns which we should modernize


acorncom (approved):
Looks good here, thanks for your work @SergeAstapov

PR #705, opened 2025-04-26
Convert addon to v2 format

SergeAstapov:
This makes `ember-changeset` an Embroider native aka v2 addon


package.json
```diff
@@ -1,29 +1,149 @@
...
+      "optional": true
+    }
+  },
+  "packageManager": "pnpm@10.7.0",
   "pnpm": {
     "overrides": {
-      "@embroider/macros": "1.16.11"
-    }
+      "@embroider/macros": "1.16.13",
+      "validated-changeset": "1.4.0"
```
SergeAstapov, line 126: - [ ] unpin before merg

```suggestion
```

PR #709, opened 2026-03-24
generate types from source

mansona:
I noticed when trying to use one of the helpers from `ember-changeset` that there weren't any types for the individual helpers, only hand-rolled types for the index 🤔 

This this repo is (almost) on the new addon infrastructure we can just let the `addon.declarations()` call work to export the right types. I have tried to port the existing hand-rolled type signitures across using JSDoc so things **should** all be the same as before and now we no longer need to rely on hand-rolled types 🎉 

Let me know if you have any questions

ping @SergeAstapov since you have done a lot of work on this repo recently
ping @bartocc since you worked on the original types


mansona:
@coderabbitai pause

bartocc:
This will be a great addition @mansona 🙏

SergeAstapov:
@mansona I'm totally in favor of having types. Question only - how good those types will be compared to if we convert addon code to TypeScript.

Biggest problem I've encountered so far trying to convert this addon to TypeScript  - types for `changeset-set`/`changeset-get` helpers, which ideally should properly honor deeply nested keys, which is non trivial with my knowledge of TypeScript (may try some LLM, maybe it will do better than me)

mansona:
So I'm going to start my answer with a reminder that I'm not a fan of Typescript (even though I find myself being a Typescript consultant more and more these days 🫠), so that means that I'm going to be biased in any conversation that proposes converting anything all the way to Typescript.

That being said, I find myself more and more happy with just extracting type information from JS files (similar to what Svelte does for its codebase). Also this change doesn't stop of from later converting to TS, it just gives us missing types for things that haven't been converted yet.

Personally I would like to see this merged and then if you wanted to convert the repo to full TS later that's ok too 😂

bartocc:
I think I've heard you say that before 🤣 
> not a fan of Typescript

